PR #4, opened 2017-10-11
Cleanup kafka test server class, refactor @BeforeClass and @AfterClass 

Crim:
General clean up of Kafka Test Server class, removed some dead methods, added some utility methods to get the Kafka and Zookeeper connection strings.

Also refactored BeforeClass and AfterClass usages in tests around standing up internal kafka/zookeeper servers.  This logic got moved into a re-usable ClassRule resource to remove code duplication and standardize it.


Crim:
Not sure tests will be any faster, but just a better way to accomplish what we had.

PR #7, opened 2017-10-12
General Code Clean up & POM Optimization

Crim:
Cleans up some of our dependencies and other cruft in the POM.    Fixes a bug in persistence adapter test where it wasn't closing out the adapter at the end of the test

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,7 +87,7 @@
         <!-- Define which Curator version -->
         <curatorVersion>4.0.0</curatorVersion>
 
-        <!-- Define which Curator version -->
+        <!-- Define which Curator Test version -->
         <curatorTestVersion>2.12.0</curatorTestVersion>
 
         <!-- test toggling -->
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,7 @@
         <kafkaVersion>0.10.0.1</kafkaVersion>
 
         <!-- Define which Curator version -->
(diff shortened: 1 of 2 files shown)


Crim:
Sorry about the poor branch name -- I rebased my previous branch off of master.

pom.xml
```diff
@@ -85,9 +85,10 @@
         <kafkaVersion>0.10.0.1</kafkaVersion>
 
         <!-- Define which Curator version -->
-        <curatorVersion>2.12.0</curatorVersion>
-
+        <curatorVersion>4.0.0</curatorVersion>
```
Crim: This is the verison used as a zookeeper client in our ZK implementations.

pom.xml
```diff
@@ -85,9 +85,10 @@
...
 
+        <!-- Define which Curator Test version -->
+        <curatorTestVersion>2.12.0</curatorTestVersion>
```
Crim: This is the internal zk test server that we use in tests.  From digging around on the curator website and in JIRA tickets, as far as I can tell this is the latest release that actually works, even though there are releases published up through 4.0.0.

stanlemon: What's the scoop with this in 4.0.0? This seems really weird...

Crim: No idea.  It bombs if you move it up to any version past 2.12.0.  I'll try to dig in more tomorrow and see if I can find out more.

Issue #6, opened 2017-10-11
Upgrade Kafka-Client Dependency to 0.11.0.1


Crim:
Upgrade Kafka-Client dependency to 0.11.0.1

stanlemon:
@Crim Do we know yet the backward compatibility on this upgrade?

PR #9, opened 2017-10-14
Upgrade to Kafka 0.11.0.1

stanlemon:
Addresses #6

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
@@ -39,6 +39,7 @@ import com.salesforce.storm.spout.dynamic.kafka.deserializer.Deserializer;
 import com.salesforce.storm.spout.dynamic.persistence.PersistenceAdapter;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.clients.consumer.InvalidOffsetException;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.consumer.OffsetOutOfRangeException;
 import org.apache.kafka.common.Metric;
@@ -256,7 +257,14 @@ public class Consumer implements com.salesforce.storm.spout.dynamic.consumer.Con
             } else {
                 // We do not have an existing offset saved, so start from the head
                 getKafkaConsumer().seekToBeginning(Collections.singletonList(topicPartition));
-                offset = getKafkaConsumer().position(topicPartition) - 1;
+
+                // This preserve the 0.10 behavior where a seekToBeginning() call followed by position() on an
+                // otherwise empty partition would yield us a -1.
+                try {
+                    offset = getKafkaConsumer().position(topicPartition) - 1;
+                } catch (InvalidOffsetException ex) {
+                    offset = -1L;
+                }
 
                 logger.info(
                     "Starting at the beginning of namespace {} partition {} => offset {}",
--- a/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
@@ -478,6 +478,7 @@ public class Consumer implements com.salesforce.storm.spout.dynamic.consumer.Con
 
             // If this partition was out of range
             // we simply log an error about data loss, and skip them for now.
+            // TODO invert this if statement and call continue.
             if (outOfRangePartitions.contains(assignedTopicPartition)) {
                 // The last offset we have in our persistence layer
                 final long lastPersistedOffset = persistenceAdapter.retrieveConsumerState(
@@ -489,7 +490,7 @@ public class Consumer implements com.salesforce.storm.spout.dynamic.consumer.Con
 
                 logger.error(
                     "DATA LOSS ERROR - offset {} for partition {} was out of range, last started = {}, last persisted = {},"
-                        + " lastFinished = {}, original exception = {}",
+                    + " lastFinished = {}, original exception = {}",
                     exceptionOffset,
                     assignedConsumerPartition,
                     lastStartedOffset,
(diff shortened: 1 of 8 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -256,19 +256,12 @@ public void open(
             } else {
                 // We do not have an existing offset saved, so start from the head
                 getKafkaConsumer().seekToBeginning(Collections.singletonList(topicPartition));
-                offset = getKafkaConsumer().position(topicPartition) - 1;
```
stanlemon: This throws an exception when we call it on a topic with no offsets.

@Crim Maybe we should catch the position and behave differently rather than drop this altogether. What do you think?

src/test/java/com/salesforce/storm/spout/dynamic/kafka/ConsumerTest.java
```diff
@@ -454,8 +461,10 @@ public void testConnectWithSinglePartitionOnTopicWithNoStateSaved() {
         // Since ConsumerStateManager has no state for partition 0, we should call seekToBeginning on that partition
         verify(mockKafkaConsumer, times(1)).seekToBeginning(eq(Lists.newArrayList(kafkaTopicPartition0)));
 
-        // Verify position was asked for
-        verify(mockKafkaConsumer, times(1)).position(kafkaTopicPartition0);
+        final ConsumerRecords<byte[], byte[]> consumerRecords = makeConsumerRecords(0, 1);
+        when(mockKafkaConsumer.poll(300)).thenReturn(consumerRecords);
+
+        consumer.nextRecord();
```
stanlemon: @Crim By dropping the position() call, the first call to nextRecord() gets the state in the right spot on a reset to the beginning. This seems ok to me, yeah?  Do we really need to be persisting state before we've done a poll()?

src/test/java/com/salesforce/storm/spout/dynamic/kafka/ConsumerTest.java
```diff
@@ -272,10 +275,14 @@ public void testTimedFlushConsumerState() throws InterruptedException {
...
         when(mockKafkaConsumer.partitionsFor(eq(topicName))).thenReturn(mockPartitionInfos);
 
+        final ConsumerRecords<byte[], byte[]> consumerRecords = makeConsumerRecords(0, 1);
+        when(mockKafkaConsumer.poll(300)).thenReturn(consumerRecords);
+
         // Call constructor
         final Consumer consumer = new Consumer(mockKafkaConsumer);
         consumer.setClock(mockClock);
         consumer.open(config, getDefaultVSpoutId(), getDefaultConsumerCohortDefinition(), mockPersistenceAdapter, null);
+        consumer.nextRecord();
```
stanlemon: @Crim By dropping the position() call, the first call to nextRecord() gets the state in the right spot on a reset to the beginning. This seems ok to me, yeah? Do we really need to be persisting state before we've done a poll()?

src/test/java/com/salesforce/storm/spout/dynamic/kafka/ConsumerTest.java
```diff
@@ -454,8 +461,10 @@ public void testConnectWithSinglePartitionOnTopicWithNoStateSaved() {
         // Since ConsumerStateManager has no state for partition 0, we should call seekToBeginning on that partition
         verify(mockKafkaConsumer, times(1)).seekToBeginning(eq(Lists.newArrayList(kafkaTopicPartition0)));
 
-        // Verify position was asked for
-        verify(mockKafkaConsumer, times(1)).position(kafkaTopicPartition0);
+        final ConsumerRecords<byte[], byte[]> consumerRecords = makeConsumerRecords(0, 999);
```
stanlemon: @Crim Setting the offset here to 999 makes the test past but also fundamentally changes the nature of the test.  I think we need to agree on a direction here before I fix the remaining test failures.

stanlemon:
Seems like we have three options @Crim:
(1) We don't seek the position after a reset to the beginning anymore.
(2) We seek, but catch the exception when there are no offsets on the topic and don't store state until the first poll().
(3) We change our test so that there's always one message on the topic. Note, this doesn't solve the brand spanking new topic scenario...

I started to go down the path of (1) but in hind sight I'm thinking (2) might actually be better.

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -256,19 +256,12 @@ public void open(
...
 
                 logger.info(
                     "Starting at the beginning of namespace {} partition {} => offset {}",
                     topicPartition.topic(), topicPartition.partition(), offset
                 );
             }
-
-            // Start tracking offsets on ConsumerPartition
```
Crim: Why drop this block?

src/test/java/com/salesforce/storm/spout/dynamic/utils/KafkaTestServer.java
```diff
@@ -108,6 +108,8 @@ public void start() throws Exception {
         kafkaProperties.setProperty("auto.create.topics.enable", "true");
         kafkaProperties.setProperty("zookeeper.session.timeout.ms", "30000");
         kafkaProperties.setProperty("broker.id", "1");
+        kafkaProperties.setProperty("offsets.topic.replication.factor", "1");
+        kafkaProperties.setProperty("auto.offset.reset", "latest");
```
Crim: Is the default value if not set?

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
stanlemon: With position() failing there would be nothing to store, but that's a fair concern - maybe we can be more intuitive and check if we actually set offset to anything first.

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -464,6 +478,7 @@ private void handleOffsetOutOfRange(OffsetOutOfRangeException outOfRangeExceptio
 
             // If this partition was out of range
             // we simply log an error about data loss, and skip them for now.
+            // TODO invert this if statement and call continue.
```
Crim: Lets make this change in a followup PR to keep this PR as simple as possible

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -507,33 +522,6 @@ private void handleOffsetOutOfRange(OffsetOutOfRangeException outOfRangeExceptio
                 } else {
                     resetPartitions.add(assignedTopicPartition);
                 }
-
-                continue;
-            }
-
-            // This partition did NOT have any errors, but its possible that we "lost" some messages
```
Crim: I'd pay to ship whoever fixed this bug a 6 pack of beer in Kafka-Client 🍻

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
stanlemon: K, I want to retool what's beneath this anyhow.

stanlemon:
GitHub got in a weird state with the PR and wasn't showing the rebased commits, so I'm closing in favor of #14

PR #12, opened 2017-10-17
Clean up OutOfRangeException handling

stanlemon:
I will rebase this off master after #6 is merged.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
@@ -39,6 +39,7 @@ import com.salesforce.storm.spout.dynamic.kafka.deserializer.Deserializer;
 import com.salesforce.storm.spout.dynamic.persistence.PersistenceAdapter;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.clients.consumer.InvalidOffsetException;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.consumer.OffsetOutOfRangeException;
 import org.apache.kafka.common.Metric;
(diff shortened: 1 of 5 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -442,98 +457,43 @@ private void fillBuffer() {
...
-                    assignedTopicPartition.partition()
-                );
-                // The offset that was in the error
-                final long exceptionOffset = outOfRangeException.offsetOutOfRangePartitions().get(assignedTopicPartition);
-
-                logger.error(
-                    "DATA LOSS ERROR - offset {} for partition {} was out of range, last started = {}, last persisted = {},"
-                    + " lastFinished = {}, original exception = {}",
+        // Loop over all the partitions in this exception
+        for (final TopicPartition topicPartition : outOfRangeException.offsetOutOfRangePartitions().keySet()) {
```
Crim: Optional optimization here, You can do the following and avoid multiple calls into the map

```java
for (final Map.Entry<TopicPartition, Long>  entry : outOfRangeException.offsetOutOfRangePartitions().entrySet) {
    final TopicPartition topicPartition = entry.getKey();
    final long exceptionOffset = entry.getValue();
```

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -442,98 +457,43 @@ private void fillBuffer() {
...
-                // That means we haven't started tracking any offsets yet, we should seek to earliest on this partition
-                logger.info("Partition {} has no stored offset, resetting to earliest {}", assignedConsumerPartition, lastStartedOffset);
-                resetPartitionsToEarliest(Collections.singletonList(assignedTopicPartition));
+                partitionOffsetsManager.replaceEntry(
+                    new ConsumerPartition(topicPartition.topic(), topicPartition.partition()),
+                    endingOffset
+                );
             } else {
-                logger.info("Backtracking {} offset to {}", assignedConsumerPartition, lastStartedOffset);
-                getKafkaConsumer().seek(assignedTopicPartition, lastStartedOffset);
```
Crim: Lets add a comment here that these ones are being seeked to head or tail or whatever we decide

Crim: It looks like we call seekToEarliest() is that what we wanna do? Or seek to tail?

PR #13, opened 2017-10-17
Handle startup race condition in SidelineSpout Handler

Crim:
@stanlemon Prior to splitting Sideline stuff out from DynamicSpout, we used a latch to synchronize all of the virtual spouts to start at the same time in the SpoutCoordinator.open() call on startup.

After the refactor, SpoutCoordinator.open() gets called prior to SidelineSpoutHandler.onSpoutOpen() (where virtual spouts get added) so we no longer have a guarantee around the virtual spouts starting at the same time. Once SidelineSpoutHandler starts adding VirtualSpouts they start immediately and begin consuming.  
(Relevant code in DynamicSpout.open(): https://github.com/salesforce/storm-dynamic-spout/blob/master/src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java#L165-L186 )

We had a race condition that showed up in DynamicSpoutTest::testResumingSpoutWhileSidelinedVirtualSpoutIsActive() that showed this.  Basically before the Firehose got its filter chain updated while resuming any active sidelines, it had already started consuming.

For some reason this tended to fail/flap more often on the kafka011 branches, but I believe if we adjust the order of adding virtual spouts on start should resolve the issue.  Please sanity check this?

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
+++ b/src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
@@ -127,9 +127,6 @@ public class SidelineSpoutHandler implements SpoutHandler {
             null
         );
 
-        // Our main firehose spout instance.
-        spout.addVirtualSpout(fireHoseSpout);
-
         final String topic = (String) getSpoutConfig().get(KafkaConsumerConfig.KAFKA_TOPIC);
 
         final List<SidelineRequestIdentifier> existingRequestIds = spout.getPersistenceAdapter().listSidelineRequests();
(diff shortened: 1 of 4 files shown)


src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
```diff
@@ -127,9 +127,6 @@ public void onSpoutOpen(DynamicSpout spout, Map topologyConfig, TopologyContext
             null
         );
 
-        // Our main firehose spout instance.
```
Crim: Instead of adding the fire hose immediately, lets wait until we've applied any filters that should be active when it starts.

src/test/java/com/salesforce/storm/spout/dynamic/DynamicSpoutTest.java
```diff
@@ -1212,8 +1212,15 @@ private void validateNextTupleEmitsNothing(
         // Try a certain number of times
         final int originalSize = collector.getEmissions().size();
         for (int x = 0; x < numberOfAttempts; x++) {
+            // Call next Tuple
```
Crim: This doesn't alter the behavior of the test, just provides better debug information when the test fails

stanlemon:
Awesome, looks great @Crim!

PR #14, opened 2017-10-17
Upgrade to Kafka 0.11.0.1

stanlemon:
Rebased #9


stanlemon:
@Crim I think from a process standpoint, moving forward we should bump the POM to the next SNAPSHOT first on master, add the heading to the changelog file and then rebase our branches off that to avoid merge conflicts as we file multiple branches for a bigger release back onto master.

PR #17, opened 2017-10-17
Add checks for required configuration options

stanlemon:


stanlemon:
@Crim This bit me today, we end up with a NPE way further down. If this seems OK I'll add tests specific to these checks, just want to make sure this won't break anything else.

Crim:
Does it make more sense in here? https://github.com/salesforce/storm-dynamic-spout/blob/master/src/main/java/com/salesforce/storm/spout/dynamic/kafka/KafkaConsumerConfig.java#L134

I think we do other required field checks in SpoutConfig, so this would follow that pattern.

stanlemon:
@Crim I could go either way, what is your preference?

Crim:
:shipit:

Issue #23, opened 2017-10-21
Errors in VirtualSpouts should be reported via outputCollector.reportError()


Crim:
Currently when an error occurs within a VirtualSpout, the error is logged.  Additionally we should be reporting these errors up to Storm via the OutputCollector.reportError() method.

PR #24, opened 2017-10-21
Add ability for VirtualSpouts to report errors back to Storm

Crim:
First pass on #23 

This adds the ability for errors within Spout Coordinator / Spout Monitor to be bubbled up to storm.  We may want to continue this pattern and push access up into Virtual Spouts and/or Consumers?

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/SpoutCoordinator.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/SpoutCoordinator.java
@@ -87,7 +87,7 @@ public class SpoutCoordinator {
     private final Map<VirtualSpoutIdentifier, Queue<MessageId>> failedTuplesQueue = new ConcurrentHashMap<>();
 
     /**
-     * Buffer by spout consumer id of errors to be reported.
+     * Buffer for errors that need to be reported.
      */
     private final Queue<Throwable> reportedErrorsQueue = new ConcurrentLinkedQueue<>();
 
@@ -201,7 +201,7 @@ public class SpoutCoordinator {
      */
     private void startSpoutMonitor() {
         CompletableFuture.runAsync(spoutMonitor, getExecutor()).exceptionally((exception) -> {
-            // This fires if SpoutMonitor dies unexpectedly by throwing an exception.
+            // This fires if SpoutMonitor dies because it threw an unhandled exception.
 
             // On errors, we need to restart it.  We throttle restarts @ 10 seconds to prevent thrashing.
             logger.error("Spout monitor died unnaturally.  Will restart after 10 seconds. {}", exception.getMessage(), exception);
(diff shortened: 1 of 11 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/SpoutCoordinator.java
```diff
@@ -146,14 +152,14 @@ public boolean hasVirtualSpout(final VirtualSpoutIdentifier spoutIdentifier) {
      * Open the coordinator and begin spinning up virtual spout threads.
      * @param topologyConfig topology configuration.
      */
-    public void open(Map<String, Object> topologyConfig) {
+    public void open(final Map<String, Object> config) {
```
Crim: Renamed this argument because it was shadowing the class property topologyConfig, which could be error prone.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutMonitor.java
```diff
@@ -109,14 +114,13 @@
...
     private final MetricsRecorder metricsRecorder;
-    
+
     /**
      * Calculates progress of {@link VirtualSpout} instances.
      */
     private SpoutPartitionProgressMonitor spoutPartitionProgressMonitor;
 
     private final Map<VirtualSpoutIdentifier, SpoutRunner> spoutRunners = new ConcurrentHashMap<>();
-    private final Map<VirtualSpoutIdentifier, CompletableFuture> spoutThreads = new ConcurrentHashMap<>();
```
Crim: More details about why this is getting removed below.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutMonitor.java
```diff
@@ -203,9 +210,6 @@ public void run() {
 
             // Start monitoring loop.
             while (keepRunning()) {
-                // Look for completed tasks
```
Crim: The logic here got moved to the CompleteableFuture.handle() method below, which allows us to remove keeping a copy of the CompletableFuture (ie spoutThreads class property removed above).  It also means the code that handles when a thread ends will fire at the time the thread ends, instead of having to poll for it.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutMonitor.java
```diff
@@ -324,7 +323,7 @@ private void reportStatus() {
             executor.getCompletedTaskCount(),
             executor.getTaskCount()
         );
-        logger.info("MessageBuffer size: {}, Running VirtualSpoutIds: {}", tupleOutputQueue.size(), spoutThreads.keySet());
+        logger.info("MessageBuffer size: {}, Running VirtualSpoutIds: {}", tupleOutputQueue.size(), spoutRunners.keySet());
```
Crim: This is a no-op, just sourcing the same info from another variable.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutMonitor.java
```diff
@@ -215,15 +220,22 @@ public void run() {
...
-        } catch (Exception ex) {
-            // TODO: Should we restart the monitor?
-            logger.error("!!!!!! SpoutMonitor threw an exception {}", ex);
+        } catch (final Exception ex) {
+            // We handle restarting spout monitor in the coordinator, which is who monitors this thread.
+            // Lets report the error
+            reportError(ex);
+
+            // Log it.
+            logger.error("!!!!!! SpoutMonitor threw an exception {}", ex.getMessage(), ex);
```
stanlemon: I've never been a big fan of this log line. Probably just drop the exclamation points IMO.

PR #26, opened 2017-10-21
Revamp Log4j2 configuration into something sane

Crim:
@stanlemon We talked about this a little bit before.

This change shouldn't alter what gets logged to the console currently.  But it adds a new log file under target/logs/test.log  This log file gets all INFO level logs, excluding a few things I thought were non-relevant.

In intellJ you can adjust your default JUnit configuration to tail this log automatically.

image


Crim:
Since this is a library, I don't think we/need one outside of tests?  Whoever pulls this in would set their own config?

PR #34, opened 2017-10-24
Add DynamicSpout.removeSpout()

stanlemon:
Also moved some lingering classes around.

- [x] Add test coverage for removeSpout()
- [x] Update change log


src/main/java/com/salesforce/storm/spout/dynamic/SpoutCoordinator.java
```diff
@@ -158,6 +161,42 @@ public void addVirtualSpout(final DelegateSpout spout) {
...
+            throw new SpoutDoesNotExistException(
+                "A spout with id " + virtualSpoutIdentifier + " does not exist in the spout coordinator!",
+                virtualSpoutIdentifier
+            );
+        }
+
+        getSpoutMonitor().removeVirtualSpout(virtualSpoutIdentifier);
+
+        // Let's block until we no longer detect that the spout is in monitor.
+        while (getSpoutMonitor().hasVirtualSpout(virtualSpoutIdentifier)) {
```
stanlemon: @Crim is there a better way to do this?

src/main/java/com/salesforce/storm/spout/dynamic/SpoutCoordinator.java
```diff
@@ -158,6 +161,42 @@ public void addVirtualSpout(final DelegateSpout spout) {
...
 
     /**
+     * Remove a new VirtualSpout from the coordinator. This will signal to the monitor to request that the VirtualSpout
+     * be stopped and ultimately removed.
+     *
+     * This method will blocked until the VirtualSpout has completely stopped.
+     *
+     * @param virtualSpoutIdentifier identifier of the VirtualSpout to be removed.
+     */
+    public void removeVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifier) {
```
Crim: i wonder if this should be a checked exception?  At the least we should denote it in the javadoc and in the method definition via throws

src/main/java/com/salesforce/storm/spout/dynamic/SpoutCoordinator.java
```diff
@@ -158,6 +161,42 @@ public void addVirtualSpout(final DelegateSpout spout) {
...
+        if (!hasVirtualSpout(virtualSpoutIdentifier)) {
+            throw new SpoutDoesNotExistException(
+                "A spout with id " + virtualSpoutIdentifier + " does not exist in the spout coordinator!",
+                virtualSpoutIdentifier
+            );
+        }
+
+        getSpoutMonitor().removeVirtualSpout(virtualSpoutIdentifier);
```
Crim: I wonder if remove is the right verb here.  remove/close/kill/terminate?

src/main/java/com/salesforce/storm/spout/dynamic/SpoutCoordinator.java
Crim: So I can think of a couple ways.  You can make the method async and return a future.  Let the caller deal w/ the problem.  Or you can pass in a timeout value for how long to block for and then toss an exception if that's exceeded and let the caller deal w/ it.

Crim: Not entirely sure how a future would work in this case.  Perhaps you put this while block in the future execution?  Not sure about that one.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutMonitor.java
```diff
@@ -209,6 +209,21 @@ public boolean hasSpout(final VirtualSpoutIdentifier virtualSpoutIdentifier) {
         }
     }
 
+    /**
+     * Signals to a VirtualSpout to stop, ultimately removing it from the monitor.
+     * @param virtualSpoutIdentifier identifier of the VirtualSpout instance to request stopped.
+     */
+    public void removeVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifier) {
+        if (!hasVirtualSpout(virtualSpoutIdentifier)) {
+            throw new IllegalArgumentException(
```
Crim: probably should be consistent and toss the same exception as the other methods when the vspout id isn't found?

Crim:
lookin good so far

Issue #33, opened 2017-10-24
Trim Whitespace from Outputfields


danielsd:
I think it would improve config readability if whitespace was trimmed from the output fields list so elements can spaced out in config files. Currently whitespace is used as part of the tuple fieldname. That said I'm not too familiar with yaml conventions so if whitespace is typically excluded from list values then perhaps it's better left alone.

https://github.com/salesforce/storm-dynamic-spout/blob/26e0643fb24ee7eeffbee2dd6041e3f819081b44/src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java#L239

stanlemon:
Resolved in #35, note the preferred strategy will be to move to a list.

I will push a snapshot to maven central, should be there in a little bit.

PR #35, opened 2017-10-25
[ISSUE-33] - trim output field values after splitting

Crim:
fix for #33 

Values are now split on ',' using Gauva's split library, trimmed, and filtered for empty values.


stanlemon:
See I kind of feel like this is the wrong approach. Why don't we just make this an actual list of strings?

Crim:
See updated test case, should we still be trimming the inputs when the user defines a list?

stanlemon:
There’s no love lost from me if we drop support for the strong version altogether. Maybe not in .9, but maybe in .10? For now I think we should consider that a deprecated format for config and add warn log so that people who use that format are encouraged to switch.

stanlemon:
Oh and no to your question, we should not trim lists. They need to just do that correctly. Part of the rationale for a list is to get out of the business of reformatting people’s config.

Crim:
Yea sounds like a plan.  Will accept strings with a warning log that it's deprecated.

PR #38, opened 2017-10-26
Remove unused deserializer method.

danielsd:
Resolves https://github.com/salesforce/storm-dynamic-spout/issues/39


Crim:
Can you update the changelog with a reference to your change?

stanlemon:
@Crim before I push to maven central, I wonder if I should bump the pom to 0.9.1? Dropping that interface method is going to trickle down and break anything that implemented it (even if it was, obviously, not used).

PR #40, opened 2017-10-30
Metrics cleanup

stanlemon:


stanlemon:
FWIW @Crim I do think this is an improvement over the metrics handling in `VirtualSpout` today.

Crim:
I need to get a chance re-review this.  I think there's some good ideas here, just need to massage it a bit.

Crim:
@stanlemon give this a look over and let me know what you think.  I went back over it and realized I made a few stupid mistakes (re-implemented counter() essentially).

Its pretty verbose, and at some point we should probably remove these metrics, but for now they're kind of handy to diagnose perf issues in the critical logic loops.

src/main/java/com/salesforce/storm/spout/dynamic/metrics/StormRecorder.java
```diff
@@ -176,6 +176,11 @@ public void assignValue(final Class sourceClass, final String metricName, final
     public void timer(final Class sourceClass, final String metricName, final long timeInMs) {
         final String key = generateKey(sourceClass, metricName);
         timers.scope(key).update(timeInMs);
+
+        // Update total time value.  This tracks how much time in total has been
+        // spent in this key, in milliseconds
+        final String totalTimeKey = metricName + "_totalTimeMs";
```
Crim: This is kind of cool, now we'll record both the average time for the metric, but also the total time spent in that area during a run under a counter.

So you'll have both:
my metric ran on average within 1242.2 milliseconds, and in total spend 2342342349ms

stanlemon:
Looks good @Crim - I can't approve since I opened this PR.  I'd say update the README/CHANGELOG and lets get this merged.

Crim:
Not sure these metrics warrant being documented since they're measuring internal performance metrics really only relevant to developers, and I expect them to be removed at some point in the future.

One thing this has convinced me of is that we should probably go ahead with your idea of enumerating these metrics so we can auto-generate documentation.

Issue #42, opened 2017-10-31
[KAFKA] Limit the number of times fillBuffer() can be called recursively


stanlemon:
See https://github.com/salesforce/storm-dynamic-spout/blob/56a4cd0c0aae9a1689ab2ccbf7cc08f8e6842d2a/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java#L404

To avoid a Stackoverflow we should limit the number of times fillBuffer can be called recursively, basically setting an upper bounds to the times we attempt this during an OutOfRangeException.

PR #43, opened 2017-11-01
Limit the recursion in fillBuffer()

stanlemon:
Resolves issue #42

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
@@ -287,8 +287,9 @@ public class Consumer implements com.salesforce.storm.spout.dynamic.consumer.Con
         // Fill our buffer if its empty
         fillBuffer();
 
-        // Check our iterator for the next message
-        if (!bufferIterator.hasNext()) {
+        // Check our iterator for the next message, it's only null at this point if something bad is happening inside
+        // of the fileBuffer() method, like it bailed after too much recursion. There will be a lot for that scenario.
+        if (bufferIterator == null || !bufferIterator.hasNext()) {
             // Oh no!  No new msg found.
             logger.debug("Unable to fill buffer...nothing new!");
             return null;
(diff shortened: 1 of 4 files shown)


stanlemon:
@danielsd FYI

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -400,8 +410,17 @@ private void fillBuffer() {
                 buffer = null;
                 bufferIterator = null;
 
-                // TODO: heh... this should be bounded most likely...
-                fillBuffer();
+                // Why 5? Because it's less than 6.
+                if (trips >= 5) {
+                    logger.error(
+                        "Attempted to fill the buffer after an OffsetOutOfRangeException, but this was my fifth attempt so I'm bailing."
```
Crim: i wonder if we should do it the number of partitions + 1 or something.  Or if that's just as arbitrary as 5

Crim:
looks good to me.

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
stanlemon: There isn't a good way to determine the number of available frames are there?

Crim: partitions? i think there is, but as your comment reads, it may not much matter because nextTuple() will be called rather frequently.

Crim: The more I think about it, the more it probably just makes sense to have some sane limit, if 5 is what we have, 5 it is. 🤷‍♂️

stanlemon: I meant frames as in the stack. Remember when we hit this we saw a stack overflow. So if we didn’t want a hard limit I was postulating we could see how many frames were left in the stack. But that seems unnecessarily complicated.

PR #44, opened 2017-11-01
Add consumer id to the Record object

### LINKED ISSUE #41: Add consumer id to consumer record
The consumer id is a handy piece of information that could be used by a topology component to determine which virtual spout a tuple is associated with. Adding the consumer id to the consumer record and passing it into the deserializer's deserialize method should allow users to extend the deserializer to make use of it. This came out of a conversation with Stan so he should have more details if necessary.

stanlemon:
Resolves #41

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/consumer/Record.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/consumer/Record.java
@@ -31,6 +31,7 @@ import org.apache.storm.tuple.Values;
  * Represents the next 'Record' coming from a Consumer instance.
  */
 public class Record {
+    private final String consumerId;
     private final String namespace;
     private final int partition;
     private final long offset;
@@ -43,11 +44,16 @@ public class Record {
      * @param offset offset of the record.
      * @param values values of the record.
      */
-    public Record(final String namespace, final int partition, final long offset, final Values values) {
+    public Record(final String consumerId, final String namespace, final int partition, final long offset, final Values values) {
+        this.consumerId = consumerId;
         this.namespace = namespace;
         this.partition = partition;
         this.offset = offset;
         this.values = values;
+    }
+
+    public String getConsumerId() {
+        return consumerId;
     }
 
     public String getNamespace() {
@@ -69,7 +75,8 @@ public class Record {
     @Override
     public String toString() {
         return "Record{"
-            + "namespace='" + namespace + '\''
+            + "consumerId='" + consumerId + '\''
+            + ", namespace='" + namespace + '\''
             + ", partition=" + partition
             + ", offset=" + offset
             + ", values=" + values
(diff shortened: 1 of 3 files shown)


stanlemon:
@Crim Interestingly when it comes out of the Vspout the MessageId has this info, we just don't have it further upstream. I kind of went back and forth on this, but the potential use case is when tracking offsets like we were doing in our debug scenario and you have sidelines going. It *could* give you the ability to introspect the source of the message a little more closely inside of the Deserializer. Maybe this is over kill though, I'm not sure.  What do you think?

Crim:
I'm curious to know more about the use case here.  I'm wondering if this encourages tightly coupling downstream logic in the topology to internals of the spout, which feels "Not Good" (tm) 

On the other hand, I could see this being useful in the context of being able to emit the tuple down a different stream based on the virtual spout it originated from, but allowing that to happen is a larger issue in itself.

src/main/java/com/salesforce/storm/spout/dynamic/consumer/Record.java
```diff
@@ -43,13 +44,18 @@
      * @param offset offset of the record.
      * @param values values of the record.
      */
-    public Record(final String namespace, final int partition, final long offset, final Values values) {
+    public Record(final String consumerId, final String namespace, final int partition, final long offset, final Values values) {
```
Crim: Is this consumerId or VirtualSpout Identifier?

stanlemon: This is the consumer id. We don't pass the vspout identifier into the consumer, we just `toString()` it. I believe we erred on that path so as to allow ourselves the ability to fully split the consumer out. I think the bigger question, from my stand point, is do we expect the Consumer interface itself to ever live outside of the DynamicSpout? I'm not sure that we do...  Maybe we split Kafka off as a separate module, but does that mean the interface can be used outside of the DynamicSpout?

stanlemon:
@Crim regarding use case, I think the idea here is that when you deserialize in the consumer you could see hey, is this from a sideline or not? There's not really facilities inside of `VirtualSpout` to do that sort of thing. That said, the deserializer is Kafka-specific so maybe we're leaking concerns here, I'm not sure.

@danielsd do you have any additional thoughts on this one?

danielsd:
It does seem like if you need to know the consumer, then maybe you shouldn't be using the dynamic spout, but that's not a choice when it comes to sidelining. I'm not sure we have a use case ourselves yet, but I think it's reasonable that a topology know which tuples came from a sideline spout so they can be processed differently if necessary.

stanlemon:
Not 100% sure I agree. Part of the goal of sidelining is so that you don't have to worry what spout something originates from. The more I think about this the less I am sold that we should add this API surface area without a clear use case to back it up.  Adding this parameter requires every implementing interface to pass this data. Not sure that's a necessary requirement just yet. If we can come up with a good use case for adding this I'm game, but I think for now I'm going to sideline (har har) this PR until we do.

PR #45, opened 2017-11-02
Drop VirtualSpout.getMaxLag() and pass the MetricsRecorder to Consumer

stanlemon:
The getMaxLag() method has always felt awkward and it's another one of those situations where we are leaking our concerns across layers. I've wanted to switch this maxLag to be the endOffset and in other Consumer implementations we have a need for the ability to report metrics.  So this PR introduces the MetricsRecorder into the Consumer interface and drops the getMaxLag() metric in favor of a consumer specific one for the starting offset.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
@@ -37,17 +37,15 @@ import com.salesforce.storm.spout.dynamic.consumer.ConsumerState;
 import com.salesforce.storm.spout.dynamic.consumer.PartitionOffsetsManager;
 import com.salesforce.storm.spout.dynamic.consumer.Record;
 import com.salesforce.storm.spout.dynamic.kafka.deserializer.Deserializer;
+import com.salesforce.storm.spout.dynamic.metrics.MetricsRecorder;
 import com.salesforce.storm.spout.dynamic.persistence.PersistenceAdapter;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.InvalidOffsetException;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.consumer.OffsetOutOfRangeException;
-import org.apache.kafka.common.Metric;
-import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
-import org.apache.kafka.common.internals.Topic;
 import org.apache.storm.tuple.Values;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -60,6 +58,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 
 /**
  * A high level kafka consumer that handles state/log offset management in a way that supports
@@ -92,11 +91,20 @@ import java.util.Set;
  */
 // TODO - rename this class?
 public class Consumer implements com.salesforce.storm.spout.dynamic.consumer.Consumer {
-    // For logging.
+
+    /**
+     * Logger for logging logs.
+     */
     private static final Logger logger = LoggerFactory.getLogger(Consumer.class);
 
-    // Kafka Consumer Instance and its Config.
+    /**
+     * KafkaConsumer configuration.
+     */
     private KafkaConsumerConfig consumerConfig;
+
+    /**
+     * KafkaConsumer, for pulling messages off of a Kafka topic.
+     */
     private KafkaConsumer<byte[], byte[]> kafkaConsumer;
 
     /**
@@ -110,6 +118,11 @@ public class Consumer implements com.salesforce.storm.spout.dynamic.consumer.Con
      * loading the last known consumer state back in.
      */
     private PersistenceAdapter persistenceAdapter;
+
+    /**
+     * MetricsRecorder for reporting metrics from the consumer.
+     */
+    private MetricsRecorder metricsRecorder;
 
     /**
      * Our Deserializer, it deserializes messages from Kafka into objects.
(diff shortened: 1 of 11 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -329,6 +363,30 @@ public Record nextRecord() {
...
+        }
+
+        final Map<TopicPartition, Long> topicPartitions = getKafkaConsumer().endOffsets(
+            getKafkaConsumer().assignment()
+        );
+
+        for (final Map.Entry<TopicPartition, Long> topicPartition : topicPartitions.entrySet()) {
+            metricsRecorder.assignValue(
+                getClass(),
+                "endOffset.topic." + topicPartition.getKey().topic() + ".partition." + topicPartition.getKey().partition(),
```
Crim: the ".topic." partition in the key might not be needed

Crim:
While "maxLag" was kind of a garbage stat before, it looks like we lost it entirely in this PR?

Should we be calculating the difference between the current state and end offset for each partition and reporting that?

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
stanlemon: I suppose it's implied, but if the same topology has two instances of the same spout wouldn't that cause this to overlap?

stanlemon:
@Crim We should be able to calculate more accurate max lag using endOffset, no? Isn't that endOffset - currentOffset for each partition?

Crim:
Where is that reported in the consumer?  Or is that in SpoutPartitionProgressMonitor.java?  Seems weird to have that logic split across two un-related classes.

Does it make sense to just entirely remove SpoutPartitionProgressMonitor.java and fold that logic elsewhere?

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -329,6 +363,48 @@ public Record nextRecord() {
...
+
+        final Map<TopicPartition, Long> topicPartitions = getKafkaConsumer().endOffsets(
+            getKafkaConsumer().assignment()
+        );
+
+        // Capture the ending offset for this topic/partition on the kafka topic
+        for (final Map.Entry<TopicPartition, Long> topicPartition : topicPartitions.entrySet()) {
+            metricsRecorder.assignValue(
+                getClass(),
+                "endOffset.topic." + topicPartition.getKey().topic() + ".partition." + topicPartition.getKey().partition(),
```
stanlemon: @Crim should we do topic / partition / metric or metric / topic / partition? I can't decide...

stanlemon:
@Crim I feel like what we want/need is a set of metrics at the `DynamicSpout` and `VirtualSpout` layers that are independent of the `Consumer` metrics. Right now they are kind of convoluted, but `SpoutPartitionProgressMonitor` has the sort of thing I suspect we want, just not sure if that's the final resting place. Because that class isn't public and we're not forcing an implementing method I'm OK with it right now, but we should think about a better structure long term.

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -329,6 +363,48 @@ public Record nextRecord() {
...
+
+            if (!topicPartitions.containsKey(topicPartition)) {
+                continue;
+            }
+
+            final Long endOffset = topicPartitions.get(topicPartition);
+
+            metricsRecorder.assignValue(
+                getClass(),
+                "lag.topic." + consumerPartition.namespace() + ".partition." + consumerPartition.partition(),
```
stanlemon: @Crim Another thing I've been thinking about lately... should we consider adding an enum of metric names and String.format() these, or even change the metrics recorder to take a series of values for use in String.format() similar to how the logger does? One of the big advantages of having an enum, or a class of Strings even will be that we can generate docs for supported metrics really easily, just like we do configuration values.

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -384,6 +385,23 @@ private void reportStatus() {
...
+                endOffset - getCurrentState().getOffsetForNamespaceAndPartition(consumerPartition)
```
Crim: It'd probably be better to call getCurrentState() once and re-use the instance instead of rebuilding it for every iteration of the loop

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
Crim: Might be an interesting idea.  Perhaps part of a separate PR?

Crim:
RE: SpoutPartitionProgressMonitor. The original purpose was to abstract out the common metrics across various consumers so it wouldn't have to be re-implemented for every one, but perhaps thats an early/not needed optimization?

stanlemon:
@Crim Yeah, I think what I learned from our other consumer implementation is that the metrics we thought were generic on the kafka one weren't necessarily what I wanted to add. I think standardizing metrics in VirtualSpout about tuples passing through is ok, and I guess I'm saying I feel like that's more what SpoutPartitionProgressMonitor is doing - so it's ok.

stanlemon:
@Crim Are we good to go on this one now?

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -330,6 +364,48 @@ public Record nextRecord() {
     }
 
     /**
+     * Reports kafka consumer statistics at an interval.
+     */
+    private void reportStatus() {
+        // If we've reported status more recently than 30 seconds we skip over this.
+        if (lastReportedStatusMillis != null && TimeUnit.MILLISECONDS.toSeconds(clock.millis() - lastReportedStatusMillis) < 30) {
```
Crim: So we have a hardcoded value of 60 in StormRecorder, and a hardcoded value of 30 here.  I think for now that's fine, in the future we may want to think about making them configurable somewhere.

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
Crim: My 2 cents.  I'd do topic.partition.metric.  We may have more than one metric listed under there in the future.

Crim: Like lag below.  We can keep the keyspace lower by putting the related metrics together.

Crim:
Looks good.  I made a slight tweak in this PR: https://github.com/salesforce/storm-dynamic-spout/pull/50 which wants to merge into your PR here.  

Feel free to merge or discard.

stanlemon:
@Crim I think we should switch to String.format() in #50, but then I'm fine with you merging it into this branch.

Crim:
For readability I'd agree with you, tho the interwebs seems to believe that String.format() was considerably slower than just concat'ing

Crim:
TODO: Update README with new / removed metrics.

Crim:
I updated the README.  I think the Changelog could use some work, may tackle that next.

Crim:
Ok, seems solid to me @stanlemon 👍

PR #47, opened 2017-11-02
SidelineSpoutHandler should periodically check its state and recover

stanlemon:
VirtualSpouts can die. They shouldn't, but they do. FilterChains can get out of whack. They shouldn't, but they do. The SidelineSpoutHandler needs to periodically check its state and reload things to get them in the correct state.  This should be as simple as calling `loadSidelines()` periodically, but we need to add test coverage that jacking with the state of the spouts running and the various filter chains resumes correctly with subsequent calls. Then when all is said and done we can call loadSildelines() on an interval and we should be good to go.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
@@ -395,6 +395,14 @@ public class DynamicSpout extends BaseRichSpout {
     }
 
     /**
+     * Get the total number of virtual spouts in the coordinator.
+     * @return total number of virtual spouts in the coordinator.
+     */
+    public int getTotalVirtualSpouts() {
+        return getCoordinator().getTotalSpouts();
+    }
+
+    /**
      * @return The spout's output collector.
      */
     private SpoutOutputCollector getOutputCollector() {
--- a/src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
@@ -396,6 +396,9 @@ public class DynamicSpout extends BaseRichSpout {
 
     /**
      * Get the total number of virtual spouts in the coordinator.
+     *
+     * This method crosses the thread barrier to determine its value.
+     *
      * @return total number of virtual spouts in the coordinator.
      */
     public int getTotalVirtualSpouts() {
--- a/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
@@ -271,6 +271,8 @@ public class VirtualSpout implements DelegateSpout {
 
         startingState = null;
         endingState = null;
+
+        isOpened = false;
     }
 
     /**
--- a/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
@@ -271,8 +271,6 @@ public class VirtualSpout implements DelegateSpout {
 
         startingState = null;
         endingState = null;
-
-        isOpened = false;
     }
 
     /**
--- a/src/main/java/com/salesforce/storm/spout/sideline/config/SidelineConfig.java
+++ b/src/main/java/com/salesforce/storm/spout/sideline/config/SidelineConfig.java
@@ -51,6 +51,15 @@ public class SidelineConfig {
     public static final String TRIGGER_CLASS = "sideline.trigger_class";
 
     /**
+     * (Integer) Interval (in seconds) to check running sidelines and refresh them if necessary.
+     */
+    @Documentation(
+        description = "Interval (in seconds) to check running sidelines and refresh them if necessary.",
+        type = Integer.class
+    )
+    public static final String REFRESH_INTERVAL_SECONDS = "sideline.refresh_interval_seconds";
+
+    /**
      * (String) Defines which PersistenceAdapter implementation to use.
      * Should be a full classpath to a class that implements the PersistenceAdapter interface.
      * Default Value: "com.salesforce.storm.spout.dynamic.persistence.ZookeeperPersistenceAdapter"
@@ -172,6 +181,16 @@ public class SidelineConfig {
                 "Unspecified configuration value for {} using default value {}",
                 PERSISTENCE_ZK_RETRY_INTERVAL,
                 clonedConfig.get(PERSISTENCE_ZK_RETRY_INTERVAL)
+            );
+        }
+
+        if (!clonedConfig.containsKey(REFRESH_INTERVAL_SECONDS)) {
+            // Default to 10 minutes
+            clonedConfig.put(REFRESH_INTERVAL_SECONDS, 600);
+            logger.info(
+                "Unspecified configuration value for {} using default value {}",
+                REFRESH_INTERVAL_SECONDS,
+                clonedConfig.get(REFRESH_INTERVAL_SECONDS)
             );
         }
 
--- a/src/main/java/com/salesforce/storm/spout/sideline/config/SidelineConfig.java
+++ b/src/main/java/com/salesforce/storm/spout/sideline/config/SidelineConfig.java
@@ -54,7 +54,7 @@ public class SidelineConfig {
      * (Integer) Interval (in seconds) to check running sidelines and refresh them if necessary.
      */
     @Documentation(
-        description = "Interval (in seconds) to check running sidelines and refresh them if necessary.",
+        description = "Interval (in seconds) to check running sidelines and refresh them if necessary. Defaults to 600.",
         type = Integer.class
     )
     public static final String REFRESH_INTERVAL_SECONDS = "sideline.refresh_interval_seconds";
--- a/src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
+++ b/src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
@@ -58,6 +58,9 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.Timer;
+import java.util.TimerTask;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Handler for managing sidelines on a DynamicSpout.
@@ -153,6 +156,20 @@ public class SidelineSpoutHandler implements SpoutHandler, SidelineController {
 
         createSidelineTriggers();
 
+        final long refreshIntervalSeconds = ((Number) spoutConfig.get(SidelineConfig.REFRESH_INTERVAL_SECONDS)).longValue();
+
+        final long refreshIntervalMillis = TimeUnit.SECONDS.toMillis(refreshIntervalSeconds);
+
+        // Repeat our sidelines check periodically
+        final Timer timer = new Timer();
+        timer.scheduleAtFixedRate(new TimerTask() {
+            @Override
+            public void run() {
+                loadSidelines();
+            }
+        }, refreshIntervalMillis, refreshIntervalMillis);
+
+        // Why not just start the timer at 0? Because we want to block onSpoutOpen() until the first run of loadSidelines()
         loadSidelines();
 
         for (final SidelineTrigger sidelineTrigger : sidelineTriggers) {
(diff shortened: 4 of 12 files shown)


src/test/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandlerTest.java
```diff
@@ -449,7 +397,178 @@ public void testGenerateVirtualSpoutId() {
...
+
+        final VirtualSpoutIdentifier stopRequestVirtualSpoutIdentifier = new SidelineVirtualSpoutIdentifier(
+            CONSUMER_ID_PREFIX,
+            stopRequestId
+        );
+
+        // Reload the sidelines, this is normally done via a thread on an interval - we're going to validate it behaves correctly now
+        sidelineSpoutHandler.loadSidelines();
+
+        assertFalse(
```
Crim: perhaps asserting on a specific size is better?

src/test/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandlerTest.java
```diff
@@ -449,7 +397,178 @@ public void testGenerateVirtualSpoutId() {
...
         assertEquals(expectedSidelineRequestIdentifier, sidelineVirtualSpoutIdentifier.getSidelineRequestIdentifier());
     }
+
+    /**
+     * Test that periodically calling {@link SidelineSpoutHandler#loadSidelines()} will restore {@link VirtualSpout}'s and
+     * {@link com.salesforce.storm.spout.dynamic.filter.FilterChainStep}'s after calling
+     * {@link SidelineSpoutHandler#onSpoutOpen(DynamicSpout, Map, TopologyContext)}, and that when spouts and filters are removed
+     * they are properly restored back onto the spout.
+     */
+    @Test
```
Crim: So this tests if the vspouts are no longer there, but what if you call loadSidelines and they're still there?

stanlemon: Good call, I added another call to loadSidelines() and another set of assertions to verify nothing has changed.

stanlemon:
@Crim Are we good to go on this one now?

src/test/resources/log4j2.xml
```diff
@@ -38,6 +30,14 @@
             <AppenderRef ref="File"/>
         </Logger>
 
+        <!-- Relevant classes we want to see INFO logs on the Console -->
+        <Logger name="com.salesforce.storm.spout.dynamic">
```
stanlemon: @Crim note the change here, explicitly moving the INFO to the file rather than the console. Should make things less noisy in Travis.

Crim: Is the comment incorrect now and need to updated to match?

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
```diff
@@ -395,6 +395,14 @@ public boolean hasVirtualSpout(final VirtualSpoutIdentifier spoutIdentifier) {
     }
 
     /**
+     * Get the total number of virtual spouts in the coordinator.
+     * @return total number of virtual spouts in the coordinator.
+     */
+    public int getTotalVirtualSpouts() {
```
Crim: This is crossing the thread barrier, may want to validate that its ok/safe to do so.  We may also want to start denoting when we do this

src/main/java/com/salesforce/storm/spout/sideline/config/SidelineConfig.java
```diff
@@ -51,6 +51,15 @@
     public static final String TRIGGER_CLASS = "sideline.trigger_class";
 
     /**
+     * (Integer) Interval (in seconds) to check running sidelines and refresh them if necessary.
+     */
+    @Documentation(
```
Crim: We may want to say what the default value is here for reference.

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
```diff
@@ -153,6 +156,27 @@ public void onSpoutOpen(
...
+        }
+
+        final long refreshIntervalMillis = TimeUnit.SECONDS.toMillis(refreshIntervalSeconds);
+
+        // Repeat our sidelines check periodically
+        final Timer timer = new Timer();
+        timer.scheduleAtFixedRate(new TimerTask() {
+            @Override
+            public void run() {
+                loadSidelines();
```
Crim: You may want to put a try/catch or osmething around this.  If loadSidelines() ever throws an exception, it will never fire again after that.

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
```diff
@@ -153,6 +156,27 @@ public void onSpoutOpen(
...
+            }
+        }, refreshIntervalMillis, refreshIntervalMillis);
+
+        // Why not just start the timer at 0? Because we want to block onSpoutOpen() until the first run of loadSidelines()
         loadSidelines();
```
Crim: I think you want to run loadSidelines before you schedule the task?  You also may want to make loadSidelines() synchronized to prevent multiple threads/tasks from running it at the same time?

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
```diff
@@ -153,6 +156,27 @@ public void onSpoutOpen(
 
         createSidelineTriggers();
 
+        final long refreshIntervalSeconds;
+
+        // This should be set by SidelineConfig.setDefaults(), but just in case we want to avoid the NullPointer
+        if (spoutConfig.containsKey(SidelineConfig.REFRESH_INTERVAL_SECONDS)) {
```
Crim: I'd argue that since SpoutConfig or SidelineConfig or whatever sets a default, we should just add a precondition instead of having a second default.

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
```diff
@@ -163,8 +187,8 @@ public void onSpoutOpen(
     /**
      * Loads existing sideline requests that have been previously persisted and checks to make sure that they are running.
      */
-    private void loadSidelines() {
-        final VirtualSpoutIdentifier fireHoseIdentifier = new DefaultVirtualSpoutIdentifier(getVirtualSpoutIdPrefix() + ":" + MAIN_ID);
+    void loadSidelines() {
```
Crim: The logic in loadSidelines() seems suspect to me if we run it multiple times for the following reasons:

- I believe loadSidelines() should be marked synchronized since it can get called from multiple threads now, and we probably don't want concurrent executions of that method.
- We also need to be mindful of any variable loadSidelines() touches, and where else those variables are modified/read.  This could raise some complicated race conditions.
- It looks like loadSidelines() goes thru any "active" sidelines and attempts to re-add them, but then stops short with the hasVirtualSpout() call.  Since THAT call crosses some thread barriers we need to be careful about having consistent reads on what that's accessing.  Also it looks like that check is at the end, AFTER its modified the firehose's filter chain.  Meaning we could be adding things to the filter chain every 10 mins that a sideline is running.  We could have a weird race condition here between that being called and when a virtual spout is closing up.  Example:

Start VSpout A -> Adds FilterChain Step X to firehose
.. sometime passes
Running VSpout A -> close() -> triggers removal of the filterchain from firehose
Concurrently/interleaved run of loadSidelines() adds the filterchain back? Maybe even fires the sideline back up?

Crim:
Added some comments here.  I think there are some concerns around multi-threading in the loadSidelines() method that we need to be sure of before merging this in.

src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
```diff
@@ -271,6 +271,8 @@ public void close() {
 
         startingState = null;
         endingState = null;
+
+        isOpened = false;
```
Crim: Is this change intentional?  Do we want to allow re-opening a closed instance?  Do we / can we have any validation that a VSpout and its internal consumer is in a sane state to be re-opened?

stanlemon: It was intentional, but not for a functional need.  Seemed like something we should do. I always thing of open/close as constructor/destructor's and they should be resumable, but maybe I've been thinking of that wrong the whole time.

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
stanlemon: I don't have an issue with making this synchronized, but it's not clear to me what else you are suggesting be done here?

src/test/resources/log4j2.xml
stanlemon: Good catch.

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
stanlemon: I'm fine with that.

src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
Crim: Yea I don't think that's a safe assumption IMO.

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
Crim: So I think I mis-read part of the logic in loadSidelines() but my concern around making it synchronized still makes sense.  

I also worry about odd race conditions between that method running and the state it checks against variables modified by other threads.  I wonder if it'd be better to notify the handler when a vspout instance terminates abnormally vs polling?

So one race condition I could imagine is if a vspout is closing (normally) and loadSidelines() runs at just the right moment, it may try to restart a closing vspout?

stanlemon: It could restart the closing spout, potentially - but the new spout would be a no-op and close right after, so I don't think that's a huge concern honestly.  That closing spout is going to have a consumer id that matches the first and it's tracked state should match too.

All that said, I think you're onto something here... we should probably look at doing something more akin to an EventEmitter for a bunch of these signaling concerns. It would allow us to create hooks all over creation and solve some other concerns I have too.

src/main/java/com/salesforce/storm/spout/sideline/config/SidelineConfig.java
stanlemon: Done.

src/test/resources/log4j2.xml
stanlemon: Done.

stanlemon:
@Crim I think I've responded to everything that was actionable, let me know if I have missed anything.

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
```diff
@@ -153,18 +156,46 @@ public void onSpoutOpen(
 
         createSidelineTriggers();
 
+        Preconditions.checkArgument(
+            spoutConfig.containsKey(SidelineConfig.REFRESH_INTERVAL_SECONDS)
+            && spoutConfig.get(SidelineConfig.REFRESH_INTERVAL_SECONDS) != null,
+            "Refresh interval is required."
```
Crim: Probably best to show which config option specifically is missing.

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
```diff
@@ -153,18 +156,46 @@ public void onSpoutOpen(
...
+
+        final long refreshIntervalSeconds = ((Number) spoutConfig.get(SidelineConfig.REFRESH_INTERVAL_SECONDS)).longValue();
+
+        final long refreshIntervalMillis = TimeUnit.SECONDS.toMillis(refreshIntervalSeconds);
+
+        // Why not just start the timer at 0? Because we want to block onSpoutOpen() until the first run of loadSidelines()
         loadSidelines();
 
+        // Repeat our sidelines check periodically
+        final Timer timer = new Timer();
```
Crim: Do we want to cancel this on close?

stanlemon: Done

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
stanlemon: Done

Issue #46, opened 2017-11-02
[SIDELINE] Trigger reference implementation


stanlemon:
Create a reference implementation of `SidelineTrigger` using Zookeeper watches (it's our preferred state machine at the moment).

stanlemon:
@Crim as for a namespace for this, how does com.salesforce.storm.spout.sideline.trigger.example seem? Or should we do something else? Example makes sense, but I'm also afraid it might discourage people from using it out of the box, which is what I'm hoping they'll do.

Crim:
@stanlemon To remove the dependency on kafka consumer from Sideline, I think we should do one of the following in SidelineSpoutHandler::loadSidelines():

- Remove `final String topic = (String) getSpoutConfig().get(KafkaConsumerConfig.KAFKA_TOPIC);` and replace with its own configuration value.

or

- Store the topic/namespace on the sideline request, and load it from there.

thoughts?

stanlemon:
@Crim WIP over on #72 for that item.  Because the reference implementation is done I am going to go ahead and close this issue now.

PR #48, opened 2017-11-02
Add a reference implementation of a SidelineTrigger using Zookeeper watches

stanlemon:
Addresses #46

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -148,14 +148,23 @@
             <optional>true</optional>
         </dependency>
 
-        <!-- For simple Json Serialization -->
+        <!-- Apache Curator Recipes - Optional Dependency -->
+        <!-- Only required if you make use of the ZookeeperWatchTrigger -->
         <dependency>
-            <groupId>com.googlecode.json-simple</groupId>
-            <artifactId>json-simple</artifactId>
-            <version>1.1.1</version>
+            <groupId>org.apache.curator</groupId>
+            <artifactId>curator-recipes</artifactId>
+            <version>${curatorVersion}</version>
+            <optional>true</optional>
         </dependency>
 
-    <!-- Testing Dependencies -->
+        <!-- For JSON Serialization -->
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.8.2</version>
+        </dependency>
+
+        <!-- Testing Dependencies -->
 
         <!-- Deal with hamcrest incompatibilities...argh -->
         <dependency>
--- a/src/main/java/com/salesforce/storm/spout/dynamic/filter/StaticMessageFilter.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/filter/StaticMessageFilter.java
@@ -38,15 +38,23 @@ public class StaticMessageFilter implements FilterChainStep {
     /**
      * We need a way to make this instance unique from others, so we use a UUID.
      */
-    private final UUID uniqueId;
+    private final String id;
 
     public StaticMessageFilter() {
-        this.uniqueId = UUID.randomUUID();
+        this.id = UUID.randomUUID().toString();
+    }
+
+    public StaticMessageFilter(final String id) {
+        this.id = id;
     }
 
     @Override
     public boolean filter(Message message) {
         return true;
+    }
+
+    public String getId() {
+        return id;
     }
 
     @Override
@@ -60,18 +68,18 @@ public class StaticMessageFilter implements FilterChainStep {
 
         StaticMessageFilter that = (StaticMessageFilter) other;
 
-        return uniqueId.equals(that.uniqueId);
+        return id.equals(that.id);
     }
 
     @Override
     public int hashCode() {
-        return uniqueId.hashCode();
+        return id.hashCode();
     }
 
     @Override
     public String toString() {
         return "StaticMessageFilter{"
-            + "uniqueId=" + uniqueId
+            + "id=" + id
             + '}';
     }
 }
--- a/src/main/java/com/salesforce/storm/spout/dynamic/filter/StaticMessageFilter.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/filter/StaticMessageFilter.java
@@ -78,8 +78,6 @@ public class StaticMessageFilter implements FilterChainStep {
 
     @Override
     public String toString() {
-        return "StaticMessageFilter{"
-            + "id=" + id
-            + '}';
+        return id;
     }
 }
--- a/src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorHelper.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorHelper.java
@@ -26,14 +26,16 @@
 package com.salesforce.storm.spout.dynamic.persistence.zookeeper;
 
 import com.google.common.base.Charsets;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.imps.CuratorFrameworkState;
 import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.KeeperException;
-import org.json.simple.JSONValue;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -53,6 +55,11 @@ public class CuratorHelper {
     private CuratorFramework curator;
 
     /**
+     * JSON parser.
+     */
+    private final Gson gson = new GsonBuilder().create();
+
+    /**
      * Helper methods for common tasks when working with Curator.
      * @param curator curator instance.
      */
@@ -68,7 +75,7 @@ public class CuratorHelper {
      */
     public void writeJson(String path, Map data) {
         logger.debug("Zookeeper Writing {} the data {}", path, data.toString());
-        writeBytes(path, JSONValue.toJSONString(data).getBytes(Charsets.UTF_8));
+        writeBytes(path, gson.toJson(data).getBytes(Charsets.UTF_8));
     }
 
     /**
@@ -84,7 +91,10 @@ public class CuratorHelper {
             if (bytes == null) {
                 return null;
             }
-            return (Map<K, V>) JSONValue.parse(new String(bytes, Charsets.UTF_8));
+            final Gson gson = new GsonBuilder()
+                .setDateFormat("yyyy-MM-dd HH:mm:ss")
+                .create();
+            return (Map<K, V>) gson.fromJson(new String(bytes, Charsets.UTF_8), HashMap.class);
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
--- a/src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorHelper.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorHelper.java
@@ -91,9 +91,6 @@ public class CuratorHelper {
             if (bytes == null) {
                 return null;
             }
-            final Gson gson = new GsonBuilder()
-                .setDateFormat("yyyy-MM-dd HH:mm:ss")
-                .create();
             return (Map<K, V>) gson.fromJson(new String(bytes, Charsets.UTF_8), HashMap.class);
         } catch (Exception e) {
             throw new RuntimeException(e);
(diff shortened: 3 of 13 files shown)


pom.xml
```diff
@@ -148,14 +148,23 @@
             <optional>true</optional>
         </dependency>
 
-        <!-- For simple Json Serialization -->
+        <!-- Apache Curator Recipes - Optional Dependency -->
+        <!-- Only required if you make use of the ZookeeperWatchTrigger -->
         <dependency>
-            <groupId>com.googlecode.json-simple</groupId>
-            <artifactId>json-simple</artifactId>
```
stanlemon: Swapping this to Gson because Gson lets us deserialize directly to a class, where this does not.

src/main/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTrigger.java
```diff
@@ -0,0 +1,374 @@
...
+        } catch (Exception e) {
+            logger.error("Unable to parse trigger event {} {}", json, e);
+            return null;
+        }
+    }
+
+    /**
+     * Watch implementation for the sideline trigger node in Zookeeper.
+     */
+    private class SidelineTriggerWatch implements PathChildrenCacheListener {
```
stanlemon: Intentionally not making this static, so I can easily reference methods on the outer class.

stanlemon:
@Crim This one should be ready for review now.

stanlemon:
@Crim Thinking this through... we probably need to cleanup the state in Zookeeper, otherwise we will attempt to re-start and re-stop sidelines on every load. That shouldn't yield in duplicate sideline spouts, but I'm not sure we want to do that.  Maybe it's OK. At the very least we should probably delete the records once the STOP requests have been processed. What do you think?

src/main/java/com/salesforce/storm/spout/dynamic/filter/StaticMessageFilter.java
```diff
@@ -60,18 +68,18 @@ public boolean equals(Object other) {
...
     }
 
     @Override
     public int hashCode() {
-        return uniqueId.hashCode();
+        return id.hashCode();
     }
 
     @Override
     public String toString() {
```
Crim: I wonder if toString should just return the id here?

src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorHelper.java
```diff
@@ -84,7 +91,10 @@ public void writeJson(String path, Map data) {
             if (bytes == null) {
                 return null;
             }
-            return (Map<K, V>) JSONValue.parse(new String(bytes, Charsets.UTF_8));
+            final Gson gson = new GsonBuilder()
```
Crim: Seems weird that we have two Gson instances in this class.  Maybe we should just create the appropriate one in the constructor?

src/main/java/com/salesforce/storm/spout/sideline/persistence/ZookeeperPersistenceAdapter.java
```diff
@@ -177,15 +178,15 @@ public SidelinePayload retrieveSidelineRequest(SidelineRequestIdentifier id, int
 
         final FilterChainStep step = parseJsonToFilterChainSteps(json);
 
-        final Long startingOffset = (Long) json.get("startingOffset");
-        final Long endingOffset = (Long) json.get("endingOffset");
+        final Double startingOffset = (Double) json.get("startingOffset");
```
Crim: This feels weird because we use long's everywhere else.  I believe Long is the preferred way to store large integers, where Double can allow you to store LARGER numbers but with less precision (seems bad)

Crim: Is this an artifact of using Gson?  Perhaps configuring GSON with:

```java
Gson gson = new GsonBuilder().
        registerTypeAdapter(Double.class,  new JsonSerializer<Double>() {   
       .....implementation returns a Long here...
});
```

Tho even that feels like a hack.

Crim:
Makes sense to delete, or move them to a completed subnode or something.

Crim:
Would now be a good time to move this bit into a separate submodule?  sideline-zookeeper-trigger or something?

Crim:
Somewhat related, indirectly to this pr, I noticed that ZookeeperPersistenceAdapter implements Serializable, but has non-serializable class property CuratorHelper.  I think we need to make sure CuratorHelper is serializable.

This applies to both the Sideline and dynamic spout instance.

src/main/java/com/salesforce/storm/spout/sideline/persistence/ZookeeperPersistenceAdapter.java
stanlemon: This is an artifact of switching to Gson and serializing to a Map. I'd like to create a concrete class for this payload, but I think I'd prefer to do that in a separate PR if that's agreeable.

stanlemon:
I'm of the opinion we should remove the Serializable interface from the ZookeeperPersistenceAdapter. We don't actually need it, I think it snuck on over time.

Regarding deletions, the more I think about this the more I'm inclined to add an `isProcessed` boolean to the `TriggerEvent` and flip it after the sideline has been started/stopped, and leave "cleanup" to the implementer.  I've noticed that today when we use this and we do STOP requests you can't see them from the UI and that's not ideal.

src/main/java/com/salesforce/storm/spout/sideline/persistence/ZookeeperPersistenceAdapter.java
Crim: Create an issue for it so we can capture it and I'll +1 this.

Crim:
Sounds good.  Drop the Serializable marker from those classes and maybe the GSON thing and this should be good to go.

src/main/java/com/salesforce/storm/spout/dynamic/filter/StaticMessageFilter.java
stanlemon: Done, probably not a big deal honestly as this is just a test class.

src/main/java/com/salesforce/storm/spout/sideline/persistence/ZookeeperPersistenceAdapter.java
stanlemon: See #54

stanlemon:
@Crim Updated with PR feedback, and I went ahead and added the 'processed' bit to the event. I think this is ready, let me know.

src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorHelper.java
```diff
@@ -66,25 +74,42 @@ public CuratorHelper(final CuratorFramework curator) {
...
+    /**
+     * Read data from Zookeeper that has been stored as JSON.
+     *
+     * This method will return the JSON deserialized to the provided class.
+     *
+     * @param path node containing JSON to read from.
+     * @param clazz class of the object the JSON should be deserialized to.
+     * @return map representing the JSON stored within the zookeeper node.
+     */
+    public <T> T readJson(final String path, final Class<T> clazz) {
```
Crim: should this be public or private?

src/main/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTrigger.java
```diff
@@ -0,0 +1,417 @@
...
+
+                final SidelineTriggerWatch watch = new SidelineTriggerWatch();
+
+                // Now setup our watch so that we see future changes as they come through
+                final PathChildrenCache cache = new PathChildrenCache(curator, root, true);
+                cache.start(PathChildrenCache.StartMode.POST_INITIALIZED_EVENT);
+                cache.getListenable().addListener(watch);
+
+                // Block the process until we have received our initialization event.
+                while (!watch.isInitialized()) {
```
Crim: should we time this out?

src/main/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTrigger.java
```diff
@@ -0,0 +1,417 @@
...
+
+        if (triggerEvent.isProcessed()) {
+            logger.info("TriggerEvent has already been processed, skipping. {}", triggerEvent);
+            return;
+        }
+
+        final SidelineRequest sidelineRequest = buildSidelineRequest(triggerEvent);
+
+        if (sidelineRequest == null) {
+            logger.warn("Unable to build SidelineRequest from TriggerEvent {}", triggerEvent);
```
Crim: This seems like it should be an error?

src/main/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTrigger.java
```diff
@@ -0,0 +1,417 @@
...
+     * @throws NoSuchAlgorithmException Your java install is whack yo, it's missing MD5, for realz???
+     */
+    private SidelineRequestIdentifier generateSidelineRequestIdentifier(
+        final TriggerEvent triggerEvent,
+        final FilterChainStep step
+    ) throws NoSuchAlgorithmException, UnsupportedEncodingException {
+        final String json = gson.toJson(step);
+
+        final StringBuilder identifier = new StringBuilder(
+            DatatypeConverter.printHexBinary(MessageDigest.getInstance("MD5").digest(json.getBytes("UTF-8")))
```
Crim: probably ok, but isnt MD5 considered dead?

src/test/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTriggerTest.java
```diff
@@ -0,0 +1,335 @@
...
+        final Map<String,Object> startData = new HashMap<>();
+        startData.put("id", "start");
+        final TriggerEvent startTriggerEvent = new TriggerEvent(
+            SidelineType.START,
+            startData,
+            new Date(),
+            CREATED_BY,
+            DESCRIPTION,
+            false,
+            new Date()
```
Crim: Do we really wanna use java.util.Date?  Its pretty awful :/

Crim:
Looks good, made a few comments for your consideration, but feel free to roll w/ what you got.

src/test/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTriggerTest.java
stanlemon: What do you want to use? I thought you stripped joda out?

src/main/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTrigger.java
stanlemon: Dead in what way?

src/test/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTriggerTest.java
Crim: java.util.Date just has the world worst API.  Maybe the new Java8 replacement for it?

http://www.baeldung.com/java-8-date-time-intro  using GMT/UTC timezone of course :p

src/main/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTrigger.java
Crim: I think mostly for crypto / being able to easily force collisions.  In this context it's probably safe to use.

src/test/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTriggerTest.java
stanlemon: Switched out for LocalDateTime. We don't actually use these for computations, so I'm not sure how much we care about the timezone. If you think these all need to be Zoned let me know and I'll pick it up in a subsequent PR.

src/main/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTrigger.java
stanlemon: It's simple and fast and should be fairly accurate here so I'm not horribly worried. We can always make it configurable if it becomes an issue.

src/main/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTrigger.java
stanlemon: Updated.

PR #50, opened 2017-11-06
Slightly tweak how the metrics are calculated, condense into a single loop

Crim:
This PR slightly tweaks your metric collection loops in the following way:

- It condenses the two loops into a single loop, 
- Renames some of the variables to be more explicit
- Re-uses the result of getCurrentState() instead of calling it once per loop iteration
- Adds "currentOffset" metric
- Renames the metric keyspace "topic.[topic].partition.[partition].[metricName]" so its easier to browse related metrics in something like grafana/graphite.
- Updates changelog <-- May want to review and make any other changes here that I missed.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
@@ -365,6 +365,10 @@ public class Consumer implements com.salesforce.storm.spout.dynamic.consumer.Con
 
     /**
      * Reports kafka consumer statistics at an interval.
+     * Currently this reports
+     *   - "endOffset" or TAIL position of the kafka topic/partition.
+     *   - "currentOffset" or the position the consumer is currently per topic/partition.
+     *   - "lag" or difference between currentOffset and endOffset position.
      */
     private void reportStatus() {
         // If we've reported status more recently than 30 seconds we skip over this.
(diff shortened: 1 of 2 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -751,4 +771,16 @@ public void removeConsumerState() {
...
+    /**
+     * Utility method to reduce duplication around generating metric keyspaces for
+     * topic/partition metrics.
+     * @param topic Name of topic the metric belongs to.
+     * @param partition Partition the metric belongs to.
+     * @param metric Name of the metric.
+     * @return Properly formatted keyspace for metric.
+     */
+    private String getTopicPartitionMetricKey(final String topic, final int partition, final String metric) {
+        return "topic." + topic + ".partition." + partition + "." + metric;
```
stanlemon: I think we should use String.format() here.

stanlemon:
@Crim should we drop the other currentOffset metric then?

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
Crim: For readability I 100% agree.  For performance, some guy somewhere out on the internet claims its significantly slower compared to just straight concatenation.  

Unsure the best approach here.  Going to merge this in as is to your branch, and will toy with it in your branch.

Crim:
I considered it, I think for now lets just leave it.  I think we should talk at some point in which (if any) of the SpoutProgress metrics are useful.  I know you hate that class and want it to die, and I'm starting to agree with you :p

PR #55, opened 2017-11-10
Coordinator Refactor

Crim:
This is the follow on to #53 and while functionally "complete", should be considered a work in progress.  The way the PR is structured, it includes the change set from #53 which makes this look more complex than it really is.

Since #53 made SpoutCoordinator essentially pass-through class to SpoutMonitor, this PR attempts to merge the two.  Because most of the logic we wanted to keep lived in SpoutMonitor, I choose to remove SpoutCoordinator.  We should consider renaming SpoutMonitor to something else, either SpoutCoordinator, SpoutLauncher, VirtualSpoutMonitor, etc.. as well as rename its package namespace.

#### What was left in SpoutCoordinator
The core logic left in SpoutCoordinator revolved around creating a FixedSingleThreadPool ExecutorService.  Using this service we submitted SpoutMonitor to start running within it.  If SpoutMonitor ever threw an exception, SpoutCoordinator would simply log an error and re-submit it.  Every other method simply delegated/passed thru to SpoutMonitor.

#### What changed in SpoutMonitor
With removing SpoutCoordinator, I simply re-arranged SpoutMonitor such that it will now catch and log/report any exceptions and just continue to run.  Previously SpoutCoordinator was a watch dog for this process, and with it gone we want to prevent this thread from dying.  The end result should be identical as before. 

#### What changed in DynamicSpout
SpoutMonitor is now created/started/closed directly from DynamicSpout.  Calling addVirtualSpout/removeVirtualSpout/hasVirtualSpout now passes directly through to SpoutMonitor.

**Note** It may be that instead of passing the entire Spout to Spout handlers for dealing with adding/removing VirtualSpouts, we pass SpoutMonitor instead?  If we did that we can remove the add/remove/has methods from DynamicSpout which would be nice.  Unfortunately I'm not super up to speed on how the handlers work to say if this really makes sense or not, but wanted to throw it out there.


Crim:
I think I broke something in this branch/PR.  Travis keeps failing on a compilation error, however locally I have no such issues.

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
```diff
@@ -155,16 +164,24 @@ public void open(Map topologyConfig, TopologyContext topologyContext, SpoutOutpu
...
         messageBuffer.open(getSpoutConfig());
 
-        // Create Spout Coordinator.
-        coordinator = new SpoutCoordinator(
-            // Our metrics recorder.
-            metricsRecorder,
-            // Our MessageBuffer/Queue Implementation.
-            messageBuffer
+        // Create MessageBus instance and store into SpoutMessageBus reference reducing accessible scope.
+        final MessageBus messageBus = new MessageBus(messageBuffer);
```
stanlemon: LOL, this feels like cheating - but OK.

src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorFactory.java
```diff
@@ -83,13 +85,21 @@ public static CuratorFramework createNewCuratorInstance(final Map<String, Object
             .collect(Collectors.joining(","));
 
         try {
+            // Create new ThreadFactory with named threads.
+            // TODO allow pushing in better naming.
```
stanlemon: Not sure that's necessary honestly.

src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorFactory.java
```diff
@@ -83,13 +85,21 @@ public static CuratorFramework createNewCuratorInstance(final Map<String, Object
...
+            final ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                .setNameFormat("[DynamicSpout:PersistenceAdapter] Curator Pool %d")
```
stanlemon: This is going to be confusing, if you're using the reference impl there are three uses of this with three distinct curator instances. Only one of them is actually the DS-PA.  I would say let's either add a String as a first param to this method or add a name to the config object.

src/test/java/com/salesforce/storm/spout/dynamic/DynamicSpoutTest.java
```diff
@@ -1002,10 +1001,10 @@ public void testConsumeWithConsumerGroupOddNumberOfPartitions(final int taskInde
     @Test
     public void testReportErrors() {
         // Define config
-        //final Map<String, Object> config = getDefaultConfig("ConsumerIdPrefix", "StreamId");
-        final Map<String, Object> config = new HashMap<>();
+        Map<String, Object> config = new HashMap<>();
         config.put(SpoutConfig.VIRTUAL_SPOUT_ID_PREFIX, "ConsumerIdPrefix");
         config.put(SpoutConfig.PERSISTENCE_ADAPTER_CLASS, InMemoryPersistenceAdapter.class.getName());
+        config = SpoutConfig.setDefaults(config);
```
stanlemon: Elsewhere we seem to normally set defaults when creating the map and then putting onto it.

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
```diff
@@ -185,18 +202,19 @@ public void open(Map topologyConfig, TopologyContext topologyContext, SpoutOutpu
...
             .getErrors()
             .ifPresent(throwable -> getOutputCollector().reportError(throwable));
 
-        // Ask the SpoutCoordinator for the next message that should be emitted. If it returns null, then there's
+        // Ask the MessageBus for the next message that should be emitted. If it returns null, then there's
         // nothing new to emit! If a Message object is returned, it contains the appropriately mapped MessageId and
         // Values for the tuple that should be emitted.
-        final Message message = getCoordinator().nextMessage();
-        if (message == null) {
+        final Optional<Message> messageOptional = getMessageBus().nextMessage();
```
stanlemon: Optionals feel unnecessary here, I would rather we chuck check for null in this case.

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
```diff
@@ -420,9 +451,17 @@ public MetricsRecorder getMetricsRecorder() {
     /**
      * @return The virtual spout coordinator.
      */
-    SpoutCoordinator getCoordinator() {
+    SpoutCoordinator getSpoutCoordinator() {
```
stanlemon: Interestingly the only time we call this is to get the number of spouts, `SpoutCoordinator.getTotalSpouts()`.

I'd like to see SpoutCoordinator NOT be public.  I think we should do one of two things...
1. Add a `DynamicSpout.getTotalVitualSpouts()` method that basically does what we need in the tests that call this method.
2. Add a `DynamicSpout.getVirtualSpouts()` which we can call `size()` on in the tests.  In the past I've debated adding this but the complexity with the monitor -> coordinator made me avoid doing so. With as simple as this is not it actually might be worth it.

If you want we can do that in a separate PR, just write a ticket up for it.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
```diff
@@ -188,143 +166,203 @@ public SpoutMonitor(
...
+     * managed with teh other currently running spouts.
+     *
+     * This method is asynchronous.  After calling this method you simply know that the VirtualSpout has been
+     * queued to start.  There is no garuntee that it has actually started after calling this method.
+     *
+     * @param spout New delegate spout
+     * @throws SpoutAlreadyExistsException if a spout already exists with the same VirtualSpoutIdentifier.
+     */
+    public synchronized void addVirtualSpout(final DelegateSpout spout) throws SpoutAlreadyExistsException {
+
```
stanlemon: Remove this extra space.

src/main/java/com/salesforce/storm/spout/dynamic/MessageBus.java
```diff
@@ -0,0 +1,177 @@
...
+    // SpoutMessageBus Interface methods.
+
+    @Override
+    public Optional<Throwable> getErrors() {
+        // Poll is non-blocking.
+        return Optional.ofNullable(reportedErrorsQueue.poll());
+    }
+
+    @Override
+    public Optional<Message> nextMessage() {
```
stanlemon: As mentioned elsewhere, let's lose the Optionals.

src/main/java/com/salesforce/storm/spout/dynamic/MessageBus.java
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * Queue for tuples that are ready to be emitted out into the topology.
+     */
+    private final MessageBuffer messageBuffer;
+
+    /**
+     * Buffer by spout consumer id of messages that have been acked.
+     */
+    private final Map<VirtualSpoutIdentifier, Queue<MessageId>> ackedTuplesQueue = new ConcurrentHashMap<>();
```
stanlemon: This is petty, but I don't think these need to have `Queue` in the name honestly.

src/main/java/com/salesforce/storm/spout/dynamic/MessageBus.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    @Override
+    public Optional<Message> nextMessage() {
+        return Optional.ofNullable(messageBuffer.poll());
+    }
+
+    @Override
+    public void ack(final MessageId id) {
+        // TODO this actually isn't thread safe :(
```
stanlemon: I don't think this comment is true.  `containsKey` calls `get` and check that it's not `null`, so functionally it should be as thread safe as what you're doing just four or five lines down.

src/main/java/com/salesforce/storm/spout/dynamic/MessageBus.java
```diff
@@ -0,0 +1,177 @@
...
+            logger.warn("Acking tuple for unknown consumer");
+            return;
+        }
+
+        ackedTuplesQueue.get(id.getSrcVirtualSpoutId()).add(id);
+    }
+
+    @Override
+    public void fail(final MessageId id) {
+        // TODO this actually isn't thread safe :(
```
stanlemon: Same with this one, I think this is thread safe unless you're seeing something I am not.

src/main/java/com/salesforce/storm/spout/dynamic/MessageBus.java
```diff
@@ -0,0 +1,177 @@
...
+        if (!failedTuplesQueue.containsKey(id.getSrcVirtualSpoutId())) {
+            logger.warn("Failing tuple for unknown consumer");
+            return;
+        }
+
+        failedTuplesQueue.get(id.getSrcVirtualSpoutId()).add(id);
+    }
+
+    // VirtualSpoutMessageBus interface methods.
```
stanlemon: Lose this comment, this is what the annotations are for.

src/main/java/com/salesforce/storm/spout/dynamic/MessageBus.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    @Override
+    public void unregisterVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifier) {
+        messageBuffer.removeVirtualSpoutId(virtualSpoutIdentifier);
+        ackedTuplesQueue.remove(virtualSpoutIdentifier);
+        failedTuplesQueue.remove(virtualSpoutIdentifier);
+    }
+
+    // Additional Helper methods not exposed via interfaces.
```
stanlemon: Unnecessary comment

src/main/java/com/salesforce/storm/spout/dynamic/MessageBus.java
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * @return Count of acked messageIds that exist within the bus.
+     */
+    public int ackSize() {
+        int size = 0;
+        for (final Queue queue: ackedTuplesQueue.values()) {
```
stanlemon: What happens if a new VirtualSpout is registered while in this loop?

src/main/java/com/salesforce/storm/spout/dynamic/MessageBus.java
```diff
@@ -0,0 +1,177 @@
...
+        ackedTuplesQueue.remove(virtualSpoutIdentifier);
+        failedTuplesQueue.remove(virtualSpoutIdentifier);
+    }
+
+    // Additional Helper methods not exposed via interfaces.
+
+    /**
+     * @return Count of acked messageIds that exist within the bus.
+     */
+    public int ackSize() {
```
stanlemon: IMO these are only public because of a poorly written and located test. This should be package only. Same with `failSize()`

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
```diff
@@ -25,35 +25,37 @@
...
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 /**
- * Monitors and manages the lifecycle of virtual spouts.
+ * Manages the lifecycle of virtual spouts.
  */
-public class SpoutMonitor implements Runnable {
+public class SpoutCoordinator implements Runnable {
```
stanlemon: I think we need to work towards dropping the public off of this.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
```diff
@@ -359,10 +397,11 @@ private void reportStatus() {
             getNumberOfFailedTasks(),
             executor.getTaskCount()
         );
-        logger.info("MessageBuffer size: {}, Running VirtualSpoutIds: {}", tupleOutputQueue.size(), spoutRunners.keySet());
+        logger.info("MessageBuffer size: {}, Running VirtualSpoutIds: {}",
```
stanlemon: Nitpicky, after the ( drop a new line, put the ) on its own line too.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
```diff
@@ -424,8 +464,8 @@ public void close() {
         }
 
         // Loop through our runners and request stop on each
-        for (final SpoutRunner spoutRunner : spoutRunners.values()) {
-            spoutRunner.requestStop();
+        for (final SpoutContext spoutContext : runningSpouts.values()) {
```
stanlemon: Is this safe?

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutRunner.java
```diff
@@ -28,25 +28,23 @@
...
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.CountDownLatch;
+import java.util.Optional;
 
 /**
  * Manages running a VirtualSpout instance.
  * It handles all of the cross-thread communication via its Concurrent Queues data structures.
  */
-public class SpoutRunner implements Runnable {
+class SpoutRunner implements Runnable {
```
stanlemon: ❤️

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
Crim: So I could go either way on this, but I think in principle we're on the same page.

I feel like either SpoutHandlers deal w/ SpoutCoordinator to start/stop/whatever virtual spouts because it has a limited API (compared to DynamicSpout)

OR

You don't get access to SpoutCoordinator at all, and you're funneled thru DynamicSpout.

src/main/java/com/salesforce/storm/spout/dynamic/MessageBus.java
Crim: This isn't thread safe because its not atomic between does key exist and get.  Ideally we'd remove the containskey call, just call get, and look for a null value.

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
stanlemon: Right now the SpoutHandlers deal directly with the DynamicSpout. I'm definitely not opposed to changing that, but for all intents and purposes DynamicSpout is our principal API surface area.

Crim: sounds good

src/main/java/com/salesforce/storm/spout/dynamic/MessageBus.java
Crim: Not entirely sure.  ConcurrentHashmap doesn't exactly say either

```
Returns a {@link Collection} view of the values contained in this map.
     * The collection is backed by the map, so changes to the map are
     * reflected in the collection, and vice-versa.
```

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
Crim: will be difficult with it being in a subpackage from DynamicSpout unless we-rearrange the code out of that package

src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorFactory.java
Crim: Going to leave this as is for now.  Its better than it was, but sounds like it's still far from ideal.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
Crim: Should be

src/main/java/com/salesforce/storm/spout/dynamic/MessageBus.java
Crim: I 100% agree with you, but I'm unsure how to refactor the tests in a way to achieve the same behavior.  I could change them to instead Thread.sleep() for a period but that seems more likely to flap.

Crim: Going to merge this in, but if you have ideas for this comment up and I'll create a follow up PR and adjust it.

PR #56, opened 2017-11-10
Just compares the differences between CoordinatorFacades <--> CoordinatorRefactor branches

Crim:
see #55 don't merge this one.


src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
```diff
@@ -399,7 +426,54 @@ public void removeVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifi
...
+        // Create instance.
+        spoutMonitor = new SpoutMonitor(
+            topologyConfig,
+            threadContextName,
+            virtualSpoutMessageBus,
+            metricsRecorder
+        );
+
+        // Create and name thread.
+        final Thread thread = new Thread(spoutMonitor, "Spout Monitor on " + threadContextName);
```
Crim: Now if you have multiple spout instances on the same worker, the thread names will be unique like

"Spout Monitor on EventStreamSpout:0", "Spout Monitor on EventStreamSpout:1", "Spout Monitor on EventStreamSpout:2", etc..

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutMonitor.java
```diff
@@ -175,59 +182,117 @@ public boolean hasVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifi
...
+            return runningSpouts.containsKey(virtualSpoutIdentifier);
         }
     }
 
     /**
      * Signals to a VirtualSpout to stop, ultimately removing it from the monitor.
+     * This call will block waiting for the VirtualSpout instance to shutdown.
+     *
      * @param virtualSpoutIdentifier identifier of the VirtualSpout instance to request stopped.
+     * @throws SpoutAlreadyExistsException if a spout already exists with the same VirtualSpoutIdentifier.
```
Crim: poop this is the wrong exception, and the method definition is missing throws

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutRunner.java
```diff
@@ -183,7 +210,7 @@ public void run() {
      * Synchronized because this can be called from multiple threads.
```
Crim: javadoc is wrong, we use volatile for the flag.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutRunner.java
```diff
@@ -85,6 +85,33 @@
      */
     private volatile boolean requestedStop = false;
 
+    /**
+     * Constructor that does not make use of a count down latch.
```
stanlemon: Why? I hate having multiple constructors... if we don't need the latch, let's drop it. If we need it let's require it. Both is just messy IMO.

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
```diff
@@ -399,7 +426,54 @@ public void removeVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifi
...
+        thread.setDaemon(false);
+
+        // Start it.  Its intended that this thread will loop forever
+        // and as gracefully as possible handle any errors such that it never stops running
+        // unless its passed an interrupt signal or requested to stop.
+        // We have nothing watching this.
+        thread.start();
+
+        // Save reference to thread.
+        spoutMonitorThread = thread;
```
stanlemon: Why not just set this when you create the thread? Why the additional reference?

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutRunner.java
Crim: thats fair, the only place that uses it is in tests, happy to rip it out of there.

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
Crim: Set this where? on SpoutMonitor?  Seems awkward to place it there.  In reality if we don't really care about being 100% sure the thread stops on close, we don't need to keep the reference at all

stanlemon: I’m saying why bother with “thread” at all, why not store the Thread directly on spoutMonitorThread

Crim:
So I think you may be onto something here Lemon.  

Looking at what this class is doing, I don't see the reason why we should start a background thread, who's whole purpose is to start new VirtualSpouts.  With triggers de-coupled from the spout, I feel like we can just have addVirtualSpout fire it up synchronously.  This removes a fair bit of complexity (both in the multi-threaded code, and in our tests).

If we go that route and the background thread is no longer starting VirtualSpouts, that thread is now only occasionally polling for status and publishing metrics.

Crim:
TODO: go thru entire coordinator package and adjust what methods are public/private/package private

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
```diff
@@ -162,16 +167,20 @@ public void open(Map topologyConfig, TopologyContext topologyContext, SpoutOutpu
...
+            topologyContext.getThisTaskIndex()
         );
 
-        // Call open on coordinator, avoiding getter
-        coordinator.open(getSpoutConfig());
+        // Create Coordinator instance and call open.
+        spoutCoordinator = new SpoutCoordinator(
+            topologyConfig,
+            threadContext,
+            (VirtualSpoutMessageBus) messageBus,
```
stanlemon: This smells... I can't put my finger on it, but I feel like having to do this here is a sign we could this better.

Crim: I mean, its easily avoidable by using a local variable.

PR #59, opened 2017-11-13
Provide mechanism for defining metrics

Crim:
Don't really like the API that this defines, and I'm totally OK will throwing away this code.  At the very least we can take the enum defined here w/ the names, descriptions etc.. and re-work everything into something better.

The main issue is a lot of our metrics use dynamic keys, which makes pre-defining them in an ENUM more difficult.  The approach I went with is you can define your keys with {} place holders, and then pass arguments to replace the placeholders.  Perhaps @stanlemon knows of a better way to deal w/ this?


README.md
```diff
@@ -324,26 +324,6 @@ Class | Key | Type | Description
 ------|-----|------|------------
 SidelineSpout | start-sideline | Counter | How many `Start Sideline` requests have been received.
 SidelineSpout | stop-sideline | Counter | How many `Stop Sideline` requests have been received.
-VirtualSpout | `virtual-spout-id`.emit | Counter | Tuple emit count per VirtualSpout instance.
```
Crim: These would need to be auto-generated.

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
```diff
@@ -324,9 +306,6 @@ public void ack(Object id) {
 
         // Ack the tuple via the coordinator
         getCoordinator().ack(messageId);
-
-        // Update ack count metric for VirtualSpout this tuple originated from
```
Crim: got moved to the virtual spout

src/main/java/com/salesforce/storm/spout/dynamic/metrics/Metrics.java
```diff
@@ -0,0 +1,252 @@
...
+/**
+ * Enumeration of all available metrics.
+ */
+public enum Metrics implements MetricDefinition {
+
+    /*
+     * Spout Monitor Metrics.
+     */
+    @Documentation(
+        type = Documentation.Type.GAUGE,
```
stanlemon: I feel like we should probably use a different class for the type and unit attributes here.

I'd also be in favor of creating a distinct annotation for these.

src/main/java/com/salesforce/storm/spout/dynamic/metrics/Metrics.java
```diff
@@ -0,0 +1,252 @@
...
+        unit = Documentation.Unit.NUMBER,
+        category = Documentation.Category.SPOUT_MONITOR,
+        description = "Size of internal MessageBuffer."
+    )
+    spoutMonitor_bufferSize("SpoutMonitor", "bufferSize"),
```
stanlemon: This is bugging me... I kind of feel like these should be in all caps.

README.md
stanlemon: Let's revisit this part after your doc task PR.

src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
```diff
@@ -256,16 +257,12 @@ public void close() {
      */
     @Override
     public Message nextTuple() {
-        // Start total Time timer.
-        getMetricsRecorder().startTimer(getClass(), "nextTuple.entireMethod");
```
stanlemon: Not sure that I'm crazy about removing these, I think we should leave them in here honestly.

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
stanlemon: All of the metrics in this class that are scoped to `VirtualSpout` should go away. Most of them use the source spout id, I'd rather each `VirtualSpout` have it's own scoped metrics and than DynamicSpout have one that total stuff up.

stanlemon:
@Crim I rebased this and sorted through all of the merge issues. Please make sure you delete this branch locally and pull down a fresh version.

stanlemon:
@Crim I think the enums make this more challenging that it needs to be.  I've switched that to a final class that holds constants, similar to what we do for the config. I've also added `ClassMetric` and simplified `CustomMetric`, the former of which uses an actual `Class` for stricter typing and the latter that just holds a String. I'm going to stew on this some more, but I think I prefer this.  I will say I am reconsidering the use of enum's for the `Category` property on the `Documentation` annotation (same goes for the config too), because I feel like we have to couple in a way I don't want to do.

src/main/java/com/salesforce/storm/spout/dynamic/metrics/LogRecorder.java
```diff
@@ -60,13 +60,24 @@ public void close() {
...
-    public void count(Class sourceClass, String metricName) {
-        count(sourceClass, metricName, 1);
+    public void count(final MetricDefinition metric) {
+        countBy(metric, 1L, new Object[0]);
     }
 
     @Override
-    public void count(Class sourceClass, String metricName, long incrementBy) {
-        final String key = generateKey(sourceClass, metricName);
+    public void count(final MetricDefinition metric, final Object... metricParameters) {
```
stanlemon: @Crim I had to rename this because of ambiguity when compiling thanks to the varargs.

src/main/java/com/salesforce/storm/spout/dynamic/metrics/LogRecorder.java
```diff
@@ -60,13 +60,24 @@ public void close() {
...
+
+    }
+
+    @Override
+    public void countBy(final MetricDefinition metric, final long incrementBy) {
+        countBy(metric, incrementBy, new Object[0]);
+    }
+
+    @Override
+    public void countBy(final MetricDefinition metric, final long incrementBy, final Object... metricParameters) {
```
stanlemon: @Crim I had to rename this because of ambiguity when compiling thanks to the varargs.

stanlemon:
@Crim I think I'm done here... what do you think?

PR #60, opened 2017-11-13
Call close() to ensure background threads are stopped

Crim:
Noticed that after a test suite run we had a lot of extraneous threads orphaned in the background.

To find these I ended up adding code like 

```java
@AfterClass
public static void after() {
 Thread.sleep(500000); // Put a break point here, Look at the running frames in the debugger.
}
```


src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
```diff
@@ -90,7 +90,7 @@
     /**
      * Timer for periodically checking the state of the VirtualSpouts being managed.
      */
-    private final Timer timer = new Timer();
```
Crim: We could leave this final, tho I kind of like the idea of naming the threads with some contextual info.

Crim:
No worries I did the same with persistence manger tests previously.  I think somewhere there's a zk client being left behind still (not in this test class) that I have tracked down yet

PR #62, opened 2017-11-20
Fix failing tests from bad merge.

Crim:


src/test/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandlerTest.java
```diff
@@ -440,7 +434,7 @@ public void testLoadSidelines() {
         final Map<String, Object> config = getConfig();
 
         final DynamicSpout spout = new DynamicSpout(config);
-        spout.open(null, null, null);
+        spout.open(config, new MockTopologyContext(), null);
```
stanlemon: Wrong config. This is the topology config. Probably better to pass an empty HashMap if you don’t want to do null.

PR #63, opened 2017-11-20
Add missing close implementation, better naming of threads

Crim:
This PR deals with mysterious orphaned Curator threads within tests.

- Add close() definition to SidelineVirtualSpoutHandler implementation that closes Curator.
- Remove un-used SpoutHandler instance in test.
- Add context parameter to CuratorFactory.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
@@ -180,7 +180,7 @@ public class SpoutCoordinator implements Runnable {
         isOpen = true;
 
         // Create and name thread.
-        final Thread thread = new Thread(this, "[DynamicSpout:SpoutCoordinator] for " + threadContext.toString());
+        final Thread thread = new Thread(this, "[DynamicSpout:Coordinator] Monitor for " + threadContext.toString());
 
         // Mark as a User thread.
         thread.setDaemon(false);
--- a/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
@@ -147,7 +147,7 @@ public class SpoutCoordinator implements Runnable {
 
         // Create new ThreadFactory
         final String threadName = "[" + DynamicSpout.class.getSimpleName() + ":" + getClass().getSimpleName() + "] "
-            + "VirtualSpout Pool %d on " + threadContext.toString() + " ";
+            + VirtualSpout.class.getSimpleName() + " Pool %d on " + threadContext.toString() + " ";
         final ThreadFactory threadFactory = new ThreadFactoryBuilder()
             .setNameFormat(threadName)
             .setDaemon(false)
--- a/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
@@ -146,7 +146,7 @@ public class SpoutCoordinator implements Runnable {
 
         // Create new ThreadFactory
         final ThreadFactory threadFactory = new ThreadFactoryBuilder()
-            .setNameFormat("[DynamicSpout] Pool %d on " + threadContext.toString() + " ")
+            .setNameFormat("[DynamicSpout:Coordinator] Pool %d on " + threadContext.toString() + " ")
             .setDaemon(false)
             .build();
 
--- a/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
@@ -27,6 +27,7 @@ package com.salesforce.storm.spout.dynamic.coordinator;
 
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.salesforce.storm.spout.dynamic.DynamicSpout;
 import com.salesforce.storm.spout.dynamic.Tools;
 import com.salesforce.storm.spout.dynamic.VirtualSpoutMessageBus;
 import com.salesforce.storm.spout.dynamic.VirtualSpoutIdentifier;
@@ -145,8 +146,10 @@ public class SpoutCoordinator implements Runnable {
         this.threadContext = threadContext;
 
         // Create new ThreadFactory
+        final String threadName = "[" + DynamicSpout.class.getSimpleName() + ":" + getClass().getSimpleName() + "] "
+            + "VirtualSpout Pool %d on " + threadContext.toString() + " ";
         final ThreadFactory threadFactory = new ThreadFactoryBuilder()
-            .setNameFormat("[DynamicSpout:Coordinator] VirtualSpout Pool %d on " + threadContext.toString() + " ")
+            .setNameFormat(threadName)
             .setDaemon(false)
             .build();
 
@@ -180,7 +183,9 @@ public class SpoutCoordinator implements Runnable {
         isOpen = true;
 
         // Create and name thread.
-        final Thread thread = new Thread(this, "[DynamicSpout:Coordinator] Monitor for " + threadContext.toString());
+        final String threadName = "[" + DynamicSpout.class.getSimpleName() + ":" + getClass().getSimpleName() + "] "
+            + "Monitor for " + threadContext.toString();
+        final Thread thread = new Thread(this, threadName);
 
         // Mark as a User thread.
         thread.setDaemon(false);
--- a/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
@@ -146,7 +146,7 @@ public class SpoutCoordinator implements Runnable {
 
         // Create new ThreadFactory
         final ThreadFactory threadFactory = new ThreadFactoryBuilder()
-            .setNameFormat("[DynamicSpout:Coordinator] Pool %d on " + threadContext.toString() + " ")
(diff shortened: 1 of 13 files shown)


src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineVirtualSpoutHandler.java
```diff
@@ -99,4 +99,15 @@ public void onVirtualSpoutCompletion(DelegateSpout virtualSpout) {
     PersistenceAdapter getPersistenceAdapter() {
         return persistenceAdapter;
     }
+
+    /**
```
Crim: This was the biggest offender.

src/test/java/com/salesforce/storm/spout/dynamic/DynamicSpoutTest.java
```diff
@@ -565,9 +566,6 @@ public void testResumingSpoutWhileSidelinedVirtualSpoutIsActive() throws Interru
         TopologyContext topologyContext = new MockTopologyContext();
         MockSpoutOutputCollector spoutOutputCollector = new MockSpoutOutputCollector();
```
Crim: This instance was created and opened, but never used as far as I can tell, and never closed.  I opted to prune to remove it since I saw no purpose.

Crim:
@stanlemon take a look see and merge if you see fit.  I think the biggest problem was VirtualSidelineSpoutHandler was using the default close() implementation, and thus leaving Curator instances hanging around.

Just as an FYI, the number of Curator instances is just about :todamnhigh:   For the simple case of just running the Firehose, we now have a minimum of 4 Curator instances running.

1 - SidelineSpoutHandler
2 - SidelineSpoutHandler -> ZookeeperTriggerWatcher
3 - Firehose VirtualSpout
4 - SidelineVirtualSpoutHandler

Do you forsee any way we can share instances between the sideline handlers at all?  By design curator is threadsafe, and its intended to be shared across the JVM per cluster.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
```diff
@@ -146,7 +146,7 @@ public SpoutCoordinator(
 
         // Create new ThreadFactory
         final ThreadFactory threadFactory = new ThreadFactoryBuilder()
-            .setNameFormat("[DynamicSpout] Pool %d on " + threadContext.toString() + " ")
+            .setNameFormat("[DynamicSpout:Coordinator] VirtualSpout Pool %d on " + threadContext.toString() + " ")
```
stanlemon: Can we just call this SpoutCoordinator instead of DynamicSpout:Coordinator?

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
```diff
@@ -180,7 +180,7 @@ public void open() {
         isOpen = true;
 
         // Create and name thread.
-        final Thread thread = new Thread(this, "[DynamicSpout] SpoutCoordinator on " + threadContext.toString());
+        final Thread thread = new Thread(this, "[DynamicSpout:SpoutCoordinator] for " + threadContext.toString());
```
stanlemon: Same with this one.

src/main/java/com/salesforce/storm/spout/dynamic/persistence/ZookeeperPersistenceAdapter.java
```diff
@@ -89,7 +89,8 @@ public void open(Map spoutConfig) {
 
         this.curator = CuratorFactory.createNewCuratorInstance(
             // Take out spout persistence config and strip the key from it for our factory.
-            Tools.stripKeyPrefix("spout.persistence.zookeeper.", spoutConfig)
+            Tools.stripKeyPrefix("spout.persistence.zookeeper.", spoutConfig),
+            "PersistenceAdapter"
```
stanlemon: Use this.class.getSimpleName()

src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorFactory.java
```diff
@@ -88,7 +88,7 @@ public static CuratorFramework createNewCuratorInstance(final Map<String, Object
             // Create new ThreadFactory with named threads.
             // TODO allow pushing in better naming.
             final ThreadFactory threadFactory = new ThreadFactoryBuilder()
-                .setNameFormat("[DynamicSpout:PersistenceAdapter] Curator Pool %d")
+                .setNameFormat("[DynamicSpout:" + context + "] Curator Pool %d")
```
stanlemon: Instead of DynamicSpout why not Curator?

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineVirtualSpoutHandler.java
```diff
@@ -99,4 +99,15 @@ public void onVirtualSpoutCompletion(DelegateSpout virtualSpout) {
...
+     * Close the handler.
+     */
+    public void close() {
```
stanlemon: Add @Override

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineVirtualSpoutHandler.java
stanlemon: Yikes, good catch.

src/main/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTrigger.java
```diff
@@ -140,7 +139,8 @@ public void open(final Map config) {
         );
 
         curator = CuratorFactory.createNewCuratorInstance(
-            Tools.stripKeyPrefix(Config.PREFIX, config)
+            Tools.stripKeyPrefix(Config.PREFIX, config),
+            "ZookeeperWatchTrigger"
```
stanlemon: this.class.getSimpleName()

When you’re use a classes name, try to use this, it makes refactoring/renaming later sooo much easier.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
Crim: I'd prefer to keep them all with a common prefixed.  It makes it way easier to see what's spawned from our stuff when its all grouped together (sorted alphabetically) vs having to hunt through a ton of threads.

src/test/java/com/salesforce/storm/spout/dynamic/DynamicSpoutTest.java
stanlemon: Cool, probably an artifact of all our refactoring.

src/main/java/com/salesforce/storm/spout/dynamic/persistence/ZookeeperPersistenceAdapter.java
Crim: But that'll be the same regardless if its the Sideline one or the DynamicSpout one because they have the same class name.

src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorFactory.java
Crim: See reasoning above about using a common prefix

src/main/java/com/salesforce/storm/spout/dynamic/persistence/ZookeeperPersistenceAdapter.java
stanlemon: Ah, interesting...yeah I mean I guess we could prefix it before the class. Or find a way to abbreviate the package to one or two characters and do the FQN. I just prefer precision and typing on this sort of stuff.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
stanlemon: Fair enough, can we drive that off the class name then?

src/main/java/com/salesforce/storm/spout/sideline/persistence/ZookeeperPersistenceAdapter.java
```diff
@@ -100,7 +100,8 @@ public void open(final Map spoutConfig) {
 
         this.curator = CuratorFactory.createNewCuratorInstance(
             // Take out sideline persistence config and strip the key from it for our factory.
-            Tools.stripKeyPrefix("sideline.persistence.zookeeper.", spoutConfig)
+            Tools.stripKeyPrefix("sideline.persistence.zookeeper.", spoutConfig),
+            "SidelinePersistenceAdapter"
```
stanlemon: This one too

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
```diff
@@ -145,8 +146,10 @@ public SpoutCoordinator(
         this.threadContext = threadContext;
 
         // Create new ThreadFactory
+        final String threadName = "[" + DynamicSpout.class.getSimpleName() + ":" + getClass().getSimpleName() + "] "
+            + "VirtualSpout Pool %d on " + threadContext.toString() + " ";
```
stanlemon: I wonder if VirtualSpout should be a class reference here too.

src/main/java/com/salesforce/storm/spout/sideline/persistence/ZookeeperPersistenceAdapter.java
```diff
@@ -100,7 +100,8 @@ public void open(final Map spoutConfig) {
 
         this.curator = CuratorFactory.createNewCuratorInstance(
             // Take out sideline persistence config and strip the key from it for our factory.
-            Tools.stripKeyPrefix("sideline.persistence.zookeeper.", spoutConfig)
+            Tools.stripKeyPrefix("sideline.persistence.zookeeper.", spoutConfig),
+            "Sideline" + getClass().getSimpleName()
         );
```
stanlemon: SidelineSpout is a class, use that here.

Issue #19, opened 2017-10-18
Break DynamicSpout, KafkaConsumer, and SidelineSpout into separate modules


Crim:
To provide clear separation of concerns between the DynamicSpout framework, the 'Sideline Spout' implementation, and the 'Kafka' consumer implementation, we should separate these out into maven modules.

stanlemon:
@Crim I feel like the 0.9 milestone is a bit ambition, can we focus on production-readiness in 0.9 and concentrate more splitting in 0.10?

Crim:
I think after #65 the only remaining blocker for this is SpoutConfig contains a reference to Kafka Consumer to set the default consumer class.  That would need to be removed.

PR #65, opened 2017-11-21
Documentation re-arrangement

Crim:
Looking at #19 one of the required changes is how we generate our docs.  Currently the generator pulls in dependencies across what will become separate modules.  

This PR moves the Doc generating logic out, and allows each "module" to call the generator passing in the bits it needs to build its documentation.


src/main/java/com/salesforce/storm/spout/dynamic/config/DocTask.java
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Entry point into Doc generating task.
+     * @throws IOException on error writing file.
+     */
+    public static void main(final String[] args) throws IOException {
+        final Path inputPath = Paths.get("README.md");
+        final String tagArg = "DYNAMIC_SPOUT_CONFIGURATION";
+        final List<ClassSpec> classSpecs = new ArrayList<>();
+        classSpecs.add(new ClassSpec(SpoutConfig.class, SpoutConfig.setDefaults(Maps.newHashMap())));
+        classSpecs.add(new ClassSpec(KafkaConsumerConfig.class));
```
Crim: once KafkaConsumer bits get moved out to its own namespace, we'll create its own "DocTask" and remove it from being generated here.

script/generateDocs.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+## Build package
+mvn package -DskipTests=true
```
stanlemon: Why not turn this into a mvn goal or something?

src/main/java/com/salesforce/storm/spout/documentation/ClassSpec.java
```diff
@@ -0,0 +1,62 @@
...
+
+package com.salesforce.storm.spout.documentation;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Define a Class instance to generate documentation from.
+ */
+public class ClassSpec {
```
stanlemon: This seems unnecessary to me. Why not just force the classes that we want to generate docs for to use an interface that includes a setDefaults/getDefaults method?

Crim: I was just trying to change as little as possible for a first pass.  I think we need to think about how/what things like SpoutConfig actually look like

script/generateDocs.sh
Crim: Yea thought about that.  We could call out to this, or add a bunch of goals for each config type, dunno which is better

src/main/java/com/salesforce/storm/spout/documentation/ClassSpec.java
Crim: ^-- as part of a larger work item review those and come up with a plan

script/generateDocs.sh
stanlemon: I would do one goal in the root POM that takes a comma delimited list of classes or something like that.  That way it's injected at that base POM.

src/main/java/com/salesforce/storm/spout/documentation/ClassSpec.java
stanlemon: Yeah I dig ya, I just feel like adding an interface that's like `DefaultableConfig` or something would be cleaner on the implementation.

Crim: I agree.  I think we need to come up with concrete config classes instead of maps, and add that as part of it.

Crim: Going to create a follow up issue to create concrete config classes for DynamicSpout, SidelineSpout (if needed), and KafkaConsumer.

script/generateDocs.sh
Crim: Hooked this into the site goal.  IE `mvn clean site`

Didn't feel like its worth running on every compile

Crim:
So this isn't the 100% cleanest/DRY solution, but I think it satisfies what I was going for.  Removing the cross-package pollution, with now adding hooks for generating metric docs.

The README as a whole needs an overhaul ( see #8  ) but we're in a better spot than we were before the PR.  I also believe how we deal w/ Configs also needs some re-tooling ( #80 )

src/main/java/com/salesforce/storm/spout/documentation/ClassSpec.java
Crim: I looked at adding an interface but currently we use static methods to setDefaults().  So before we could apply an interface over it, we'd first need to refactor SpoutConfig / SidelineConfig into being instantiated into an instance.  

I think this is better pushed off to the specific issue/work around configs, see #80

PR #67, opened 2017-11-22
Remove metric from VirtualSpout API

stanlemon:
I want to keep the VirtualSpout (and really the DelegateSpout interface) as minimal as possible. Whenever we have a method that's public for stuff like this I think we need to re-evaluate it and consider dropping it if possible. Thus, this metric is being moved into the VirtualSpout and I am dropping it from the interface.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
@@ -256,6 +256,8 @@ public class VirtualSpout implements DelegateSpout {
      */
     @Override
     public Message nextTuple() {
+        getMetricsRecorder().assignValue(getClass(), getVirtualSpoutId() + ".filters", getFilterChain().getSteps().size());
+
         // Start total Time timer.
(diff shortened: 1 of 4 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
```diff
@@ -256,6 +256,8 @@ public void close() {
      */
     @Override
     public Message nextTuple() {
+        getMetricsRecorder().assignValue(getClass(), getVirtualSpoutId() + ".filters", getFilterChain().getSteps().size());
```
stanlemon: @Crim This seemed as good a place as any to put this, but I'm open to other suggestions.

Crim: This feels like we're adding a metric to the critical code path that won't change often.

stanlemon: Agreed, where’s a better place to stick it then?

stanlemon:
Closing in favor of #78

PR #68, opened 2017-11-22
Dropping requestStop() from the DelegateSpout

stanlemon:
I would contend that the tiers of SpoutRunner -> VirtualSpout around isStopRequested() are a bit of a smell. I think what we were doing in VirtualSpout was actually the isCompleted() method, but we hadn't exposed it.  Basically, we wanted a way for the VirtualSpout to indicate it was done processing and that it could be automatically removed from the coordinator.

Additionally, I think the addition of removeVirtualSpout() elsewhere handles the scenario of "kill this VirtualSpout early" and is a much cleaner interface for doing so.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
@@ -104,12 +104,6 @@ public class VirtualSpout implements DelegateSpout {
     private boolean isOpened = false;
 
     /**
-     * This flag is used to signal for this instance to cleanly stop.
-     * Marked as volatile because currently its accessed via multiple threads.
-     */
-    private volatile boolean requestedStop = false;
-
-    /**
      * This flag represents when we have finished consuming all that needs to be consumed.
      */
     private boolean isCompleted = false;
@@ -445,35 +439,13 @@ public class VirtualSpout implements DelegateSpout {
     }
 
     /**
-     * Call this method to request this {@link VirtualSpout} instance
-     * to cleanly stop.
+     * Whether or not this {@link VirtualSpout} has completed it's processing, which typically means that the data from {@link Consumer},
+     * specifically {@link #getCurrentState()} is now at or beyond {@link #getEndingState()}.
      *
-     * Synchronized because this can be called from multiple threads.
+     * @return true is the spout is finished processing, false if it is not.
      */
-    public void requestStop() {
-        synchronized (this) {
-            requestedStop = true;
-        }
-    }
-
-    /**
-     * Determine if anyone has requested stop on this instance.
-     * Synchronized because this can be called from multiple threads.
-     *
-     * @return - true if so, false if not.
-     */
-    public boolean isStopRequested() {
-        synchronized (this) {
-            return requestedStop || Thread.interrupted();
-        }
-    }
-
-    /**
-     * This this method to determine if the spout was marked as 'completed'.
-     * We define 'completed' meaning it reached its ending state.
-     * @return - True if 'completed', false if not.
-     */
-    private boolean isCompleted() {
+    @Override
+    public boolean isCompleted() {
         return isCompleted;
     }
 
@@ -640,9 +612,5 @@ public class VirtualSpout implements DelegateSpout {
         // Lets flip our flag to true.
         logger.info("Looks like all partitions are complete!  Lets wrap this up.");
         setCompleted();
-
-        // Cleanup consumerState.
-        // Request that we stop.
-        requestStop();
     }
 }
--- a/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutRunner.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutRunner.java
@@ -120,8 +120,11 @@ class SpoutRunner implements Runnable {
             // Record the last time we flushed.
             long lastFlush = getClock().millis();
 
-            // Loop forever until someone requests the spout to stop
-            while (!isStopRequested() && !spout.isStopRequested() && !Thread.interrupted()) {
+            // Loop forever until one of the following is true:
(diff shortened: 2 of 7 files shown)


stanlemon:
@Crim This was what I was trying to articulate the other night. I'd love some feedback on this one.

src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
```diff
@@ -447,43 +441,21 @@ public void fail(Object msgId) {
...
+    public boolean isCompleted() {
         return isCompleted;
     }
 
     /**
      * Mark this spout as 'completed.'
      * We define 'completed' meaning it reached its ending state.
      */
-    private void setCompleted() {
+    private synchronized void setCompleted() {
```
Crim: dont think this needs to be synchronized

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutRunner.java
```diff
@@ -121,21 +121,17 @@ public void run() {
...
             // Loop forever until someone requests the spout to stop
-            while (!isStopRequested() && !spout.isStopRequested() && !Thread.interrupted()) {
+            while (!isStopRequested() && !spout.isCompleted() && !Thread.interrupted()) {
                 // First look for any new tuples to be emitted.
                 final Message message = spout.nextTuple();
                 if (message != null) {
                     try {
                         getVirtualSpoutMessageBus().publishMessage(message);
                     } catch (final InterruptedException interruptedException) {
                         logger.error("Shutting down due to interruption {}", interruptedException.getMessage(), interruptedException);
```
Crim: I think you need to set Thread.interupted here and call continue to end the loop.  Or call break.   Basically someway to end the loop.  I could be wrong, but I believe when you catch the exception, the interrupted flag gets cleared.

src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
```diff
@@ -100,16 +100,10 @@
...
-     * This flag is used to signal for this instance to cleanly stop.
-     * Marked as volatile because currently its accessed via multiple threads.
-     */
-    private volatile boolean requestedStop = false;
-
     /**
      * This flag represents when we have finished consuming all that needs to be consumed.
      */
-    private boolean isCompleted = false;
+    private volatile boolean isCompleted = false;
```
Crim: Ok with leaving this volatile, but as of right now I don't think we ever read this across threads, so not sure its required.  Perhaps we do in tests.

src/test/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinatorTest.java
```diff
@@ -243,61 +243,6 @@ public void testSubmittingNewSpout() throws InterruptedException, ExecutionExcep
         assertEquals("ExecutorService has no running threads", 0, spoutCoordinator.getExecutor().getActiveCount());
     }
 
-    /**
-     * Test what happens when a running spout instance finishes "normally" by
```
Crim: Both of these tests seem relevant, just slightly changed.  The first should validate that when the spout isComplete returns true, it stops running (perhaps moved under SpoutRunnerTest).  

The 2nd test is validating that Coordinator properly queues VSpout instances, which seems like a valid thing to continue testing.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutRunner.java
stanlemon: Not sure I follow, what do you mean @Crim ?

Crim: yea

Crim: https://stackoverflow.com/questions/10401947/methods-that-clear-the-thread-interrupt-flag

src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
```diff
@@ -592,13 +564,15 @@ private void attemptToComplete() {
         // Compare it against our ending state
         for (final ConsumerPartition consumerPartition: currentState.getConsumerPartitions()) {
             // currentOffset contains the last "committed" offset our consumer has fully processed
-            final long currentOffset = currentState.getOffsetForNamespaceAndPartition(consumerPartition);
```
Crim: Curious as to what changed here in this logic that requires this swap?  I think these conditions need test cases.

Crim:
Fixed a couple things.  Made a note of a place that I think needs additional test coverage about knowing when to complete a VSpout.

Crim:
Ah this looks like the place we do that.

https://github.com/salesforce/storm-dynamic-spout/blob/lemon/request-stop/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java#L227-L237

stanlemon:
Correct @Crim and then when we cleanup sideline state we let the handler do it here https://github.com/salesforce/storm-dynamic-spout/blob/lemon/request-stop/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java#L242

Apologies it's not clear to me, did you add the additional test coverage you wanted or am I misunderstanding?

stanlemon:
@Crim ping

Crim:
I did not add test coverage in think I left a comment somewhere about it tho

stanlemon:
@Crim I backed out the conversion of the primitives to objects, which was the area you asked for an additional test case. I can no longer reproduce whatever issue I was having and don't see any reason to change it. I've also rebased this branch off of master.  Can you review this again so we can get it merged in?

PR #69, opened 2017-11-22
Allow ending state to be set after a VirtualSpout has been created

stanlemon:
This is being added so that you can create an open ended spout, meaning one where you don't yet know the ending state but might need to set it down the road.  This is the first step in moving our sideline implementation to support what's been called 'throttling'.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
@@ -503,6 +503,7 @@ public class VirtualSpout implements DelegateSpout {
      * Get the spout's current consumer state.
      * @return current consumer state
      */
+    @Override
     public ConsumerState getCurrentState() {
         // This could happen is someone tries calling this method before the vspout is opened
         if (consumer == null) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
```diff
@@ -511,14 +512,26 @@ public ConsumerState getCurrentState() {
...
         return endingState;
     }
 
+    /**
+     * Set the ending state of the {@link DelegateSpout}.for when it should be marked as complete.
+     *
+     * @param endingState ending consumer state for when the {@link DelegateSpout} should be marked as complete.
+     */
+    @Override
+    public void setEndingState(final ConsumerState endingState) {
```
Crim: should we handle if the ending state is already set?  Or just overwrite it?

src/test/java/com/salesforce/storm/spout/dynamic/VirtualSpoutTest.java
```diff
@@ -1608,6 +1608,53 @@ public void testGetCurrentState() {
...
+        );
+        virtualSpout.open();
+
+        assertNull("Ending state is not null", virtualSpout.getEndingState());
+
+        final ConsumerState endingState = ConsumerState
+            .builder()
+            .withPartition("Test", 0, 100L)
+            .build()
+        ;
```
Crim: where to have a dangling ; no?

stanlemon: I almost always dangle my semi-colons on multi-liners like this, makes it easy to comment a line out or reorder them. I can move it to the last line it if you like, just let me know.

Issue #70, opened 2017-11-22
Reconcile DelegateSpout & VirtualSpout nomenclature


stanlemon:
We're all over the place with the naming around this.  I actually am inclined to think that we should make the following name changes...

DelegateSpout => VirtualSpout
VirtualSpout => DefaultVirtualSpout

stanlemon:
@Crim debate me yo.

stanlemon:
@Crim ping, how do you feel about these renames above? If you're good to go I'm going to try and make this happen soonish.

Crim:
sure make it so :engage:

PR #71, opened 2017-11-22
Use a factory to create VirtualSpout instances

stanlemon:
IMO this is related to #70 as you can tell by some of the naming weirdness.

`FactoryManager` doesn't have any facilities for passing arguments to a constructor, which is why it was not used to create the `VirtualSpoutFactory`.

I have mixed feelings about the generics in `DelegateSpoutFactory`, in the end if this is not done we wind up casting, so I think this is a better option honestly.

Note that right now the `FilterChain` is part of `VirtualSpout` and not of `DelegateSpout`, I think this is worth calling out because it is the primary reason why someone would want to directly type against `VirtualSpout` today. Perhaps we should reconsider this, I'm open to this - we just need to make a decision that the `FilterChain` is a first class citizen in the interface and that we expect implementations to utilize this.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpoutFactory.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpoutFactory.java
@@ -1,0 +1,126 @@
+/**
+ * Copyright (c) 2017, Salesforce.com, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+ * following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this list of conditions and the following
+ *   disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+ *   disclaimer in the documentation and/or other materials provided with the distribution.
+ *
+ * * Neither the name of Salesforce.com nor the names of its contributors may be used to endorse or promote products
+ *   derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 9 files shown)


src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
```diff
@@ -221,17 +237,7 @@ synchronized void loadSidelines() {
         // If we haven't spun up a VirtualSpout yet, we create it here.
         if (fireHoseSpout == null) {
             // Create the main spout for the namespace, we'll dub it the 'firehose'
-            fireHoseSpout = new VirtualSpout(
-                // We use a normal virtual spout identifier here rather than the sideline one because this is NOT a sideline,
-                // it's our firehose.
-                fireHoseIdentifier,
-                getSpoutConfig(),
-                topologyContext,
```
Crim: I like this approach.  Do we need to store a copy of topologyContext anymore?

src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpoutFactory.java
```diff
@@ -0,0 +1,126 @@
...
+
+import java.util.Map;
+
+/**
+ * Factory for easily creating {@link DelegateSpout} instances.
+ *
+ * Handy in things like {@link com.salesforce.storm.spout.dynamic.handler.SpoutHandler} where we want to abstract away particulars, such
+ * as the things passed down from {@link DynamicSpout} such as the {@link MetricsRecorder} as an example.
+ */
+public class VirtualSpoutFactory implements DelegateSpoutFactory {
```
Crim: No test over this class?

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
stanlemon: Probably not. I don’t think we use it. Should we just drop it?

stanlemon: It's used to construct `ConsumerPeerContext`, which should maybe just be moved to a constructor parameter? Not sure.  I started stripping topologyContext out, I'll stick it in a branch in case we want to revisit but I don't think we should sweat it right now.

Crim: oh its just used in open right, i just meant dropping saving the reference in the class beyond the scope of where its used.  Not sure it matters much tho

src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpoutFactory.java
stanlemon: Added, thanks for keeping me honest.

PR #74, opened 2017-11-22
Deserialize JSON directly to SidelinePayload

### LINKED ISSUE #54: retrieveSidelineRequest() should deserialize directly to SidelinePayload
For reference see:
https://github.com/salesforce/storm-dynamic-spout/pull/48/files#diff-2023ea74ef0624d2e7668f3a80d170ecR167

Now that we've switched to Gson we can deserialize direct to an entity (yey). Additionally it might make sense to add a method like CuratorHelper.readJson(String path, Class clazz) to make this even more seamless.

stanlemon:
Resolves #54 

This ended up being a bit tricky thanks to the way that we serialize `FilterChainStep` instances.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/filter/StaticMessageFilter.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/filter/StaticMessageFilter.java
@@ -38,10 +38,9 @@ public class StaticMessageFilter implements FilterChainStep {
     /**
      * We need a way to make this instance unique from others, so we use a UUID.
      */
-    private final String id;
+    private String id = UUID.randomUUID().toString();
 
     public StaticMessageFilter() {
-        this.id = UUID.randomUUID().toString();
     }
 
     public StaticMessageFilter(final String id) {
--- a/src/test/java/com/salesforce/storm/spout/sideline/persistence/ZookeeperPersistenceAdapterTest.java
+++ b/src/test/java/com/salesforce/storm/spout/sideline/persistence/ZookeeperPersistenceAdapterTest.java
@@ -29,12 +29,15 @@ import com.google.common.base.Charsets;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import com.google.gson.Gson;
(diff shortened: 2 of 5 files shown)


src/test/java/com/salesforce/storm/spout/sideline/persistence/ZookeeperPersistenceAdapterTest.java
```diff
@@ -338,7 +357,27 @@ public void testEndToEndRequestStatePersistenceWithValidationWithIndependentZkCl
         final String storedDataStr = new String(storedDataBytes, Charsets.UTF_8);
         logger.info("Stored data string {}", storedDataStr);
         assertNotNull("Stored data string should be non-null", storedDataStr);
-        assertEquals("Got unexpected state", expectedStoredState, storedDataStr);
+        Map expectedMap = gson.fromJson(expectedStoredState, HashMap.class);
+
+        System.out.println(expectedStoredState);
```
Crim: Probably replace these with logger statements?

stanlemon: Whoops, good catch.

stanlemon: Dropped altogether during the rebase.

src/main/java/com/salesforce/storm/spout/dynamic/filter/StaticMessageFilter.java
```diff
@@ -38,10 +38,9 @@
     /**
      * We need a way to make this instance unique from others, so we use a UUID.
      */
-    private final String id;
+    private String id = UUID.randomUUID().toString();
```
Crim: why not final?

stanlemon: Something weird was happening when it deserialized when I set it inside of the constructor, I think it has to do with the way GSON reflects, but I'm not sure honestly. This worked around it.  This reminds me... this class should be moved into src/test.

PR #76, opened 2017-11-23
Remove TopologyContext from VirtualSpout

stanlemon:


stanlemon:
@Crim this is a quick, can I make it compile, type of pass.  The tests are broken right now, but this shows you what it'd look like to remove `TopologyContext` from the `VirtualSpout`.  I kind of like this because it minimizes what Storm specific details leak up from the `DynamicSpout`.

Crim:
Seems good.  I wonder if we should expand the ConsumerPeerContext object to include the component name.  As far as I know thats the only other potentially relevant info we might need.

Crim:
Yea, probably go ahead with this, I'd add the component id/name (spout id) to the object so thats available.

stanlemon:
@Crim I rebased this and made some changes to fix tests.  The build on this one should be passing now, can you re-review so I can merge it in?  One thing I noted on this... I wanted to standardize creation of VirtualSpouts inside of VirtualSpoutTest and that proved to be difficult because of the mockFactoryManager instance. I wonder if we should try to overhaul this test to use the concrete mock instances we've made?

Issue #75, opened 2017-11-22
[SIDELINE] Cleanup PersistenceAdapter.persistSidelineRequestState()


stanlemon:
- The 'id' is included in the 'request' already, this is an artifact from the past - we should remove it outright.
- The 'type' could be moved into the 'request', which probably makes more sense anyhow.
- `ConsumerPartition` + offsets could be grouped together in a single `ConsumerState` object, this would move blowing up and persisting by partition into the persistence adapter, but should not functionally change anything.  It'll cleanup the logic in the handler quite a bit too, and probably makes more sense in the adapter anyhow since the reason we do it the way we do is very specific for Zookeeper.

PR #79, opened 2017-11-23
Add support for resume state in sidelining

stanlemon:
## Summary
This PR adds a new state to the sideline sequence which we are calling **resume**. 

## Current States
Today sidelines are binary, they can be started or stopped.  Starting a sideline means that a starting offset is captured for a future virtual spout and a filter is applied to the firehose virtual spout.  This is typically done around the concept of a 'tenant', for example you might want to sideline account _Foobar_ for a period of time.  Stopping a sideline means that an ending offset is captured for a virtual spout that will be now created using the starting offset saved during the start of the sideline. That new virtual spout is immediately added to the coordinator with a filter that only emits the tenant in question and the filter is removed altogether from the firehose.

### Potential Problems
The consequences of the binary approach are that ordering gets wildly thrown off and when a sideline stops we are essentially doubling a tenants emitted messages. This can be controlled with a different MessageBuffer, but the underlying problem remains: if a tenant's operations required sidelining we go from 0-60 (or 65) almost immediately without slowing down the emission of data into the main spout.

## New States
Moving forward, and the goal of this PR is to split up the current stopping state into two different states:
1. **Resume:** This state will create a virtual spout using the starting offset from the starting state, and it will use a filter that only emits the given tenant. What's different here is that there is no ending state for the new virtual spout and the firehose still has its tenant filter.
2. **Resolve:** This state will mark the ending offset of the virtual spout for the sideline and remove the filter from the firehose.  When the resolving state is complete (fully resolved) the virtual spout for the sideline should be gone and there should be no filters on the filter chain for the tenant in question.

### Why does this matter?
The primary use we want to support here is throttling of traffic for a single tenant without having to know the ending point.  So by using this new 'resume' state and a MessageBuffer that emits at a slower rate for sidelines, someone could slow the emission of a tenant into the underlying `DynamicSpout`, and then at the point at which they are ready to resume full processing they can set that sideline to 'resolve'.

It's worth noting here that by calling resume() and resolve() back to back you should have the functional equivalent of today's stop() and that nothing is changing with the current start() process.

#### In the Future
Related to this, my suspicion is that this implementation will also give us better mechanics to control the timing of resolving the sideline so that some day in the future we can try to target the sideline once it's _near_ the tail of the firehose (probably using some tolerance level), but that's not the purpose of this PR just an additional thought on why this change makes sense.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
@@ -445,6 +445,20 @@ public class DynamicSpout extends BaseRichSpout {
     }
 
     /**
+     * Get a {@link DelegateSpout} instance from the {@link SpoutCoordinator}.
+     *
+     * This is useful is you want to manipulate the filter chain or alter the ending state after a {@link DelegateSpout} has
+     * been added to the {@link SpoutCoordinator}.
+     *
+     * @param virtualSpoutIdentifier identifier for teh {@link DelegateSpout} instance to get from the {@link SpoutCoordinator}.
+     * @return {@link DelegateSpout} instance
+     */
+    public DelegateSpout getVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifier) {
+        checkSpoutOpened();
+        return getSpoutCoordinator().getVirtualSpout(virtualSpoutIdentifier);
+    }
+
+    /**
      * Get the total number of virtual spouts in the coordinator.
      *
      * This method crosses the thread barrier to determine its value.
--- a/src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
@@ -453,7 +453,7 @@ public class DynamicSpout extends BaseRichSpout {
      * @param virtualSpoutIdentifier identifier for teh {@link DelegateSpout} instance to get from the {@link SpoutCoordinator}.
      * @return {@link DelegateSpout} instance
      */
-    public DelegateSpout getVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifier) {
+    public DelegateSpout getVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifier) throws SpoutDoesNotExistException {
         checkSpoutOpened();
         return getSpoutCoordinator().getVirtualSpout(virtualSpoutIdentifier);
     }
--- a/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
@@ -486,9 +486,9 @@ public class VirtualSpout implements DelegateSpout {
     }
 
     /**
-     * Set the ending state of the {@link DelegateSpout}.for when it should be marked as complete.
+     * Set the ending state of the {@link DelegateSpout} for when it should finish consuming.
      *
-     * @param endingState ending consumer state for when the {@link DelegateSpout} should be marked as complete.
+     * @param endingState ending consumer state for when the {@link DelegateSpout} should finish consuming.
      */
     @Override
     public void setEndingState(final ConsumerState endingState) {
--- a/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
@@ -218,6 +218,13 @@ public class SpoutCoordinator implements Runnable {
      * @return {@link DelegateSpout} instance
      */
     public synchronized DelegateSpout getVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifier) {
+        if (!hasVirtualSpout(virtualSpoutIdentifier)) {
+            throw new SpoutDoesNotExistException(
+                "VirtualSpout " + virtualSpoutIdentifier + " does not exist in the SpoutCoordinator.",
+                virtualSpoutIdentifier
+            );
+        }
+
         return runningSpouts.get(virtualSpoutIdentifier).getSpoutRunner().getSpout();
     }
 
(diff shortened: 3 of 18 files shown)


src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineController.java
```diff
@@ -38,26 +38,47 @@
...
 
     /**
-     * Stop sidelining.
+     *
+     * @param sidelineRequest
+     */
+    void resume(final SidelineRequest sidelineRequest);
+
+    /**
+     * Does a sideline exist in the finished state?
```
Crim: I agree with peek, we should figure out a better name.

stanlemon: I’m open to suggestions.

Crim: Bounded Completed Something Or oTher. 

I have no good suggestions :/

Crim:
testing 1 2 3

Crim:
22323

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
```diff
@@ -437,6 +437,11 @@ public boolean hasVirtualSpout(final VirtualSpoutIdentifier spoutIdentifier) {
         return getSpoutCoordinator().hasVirtualSpout(spoutIdentifier);
     }
 
+    public DelegateSpout getVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifier) {
```
Crim: jdocs?

src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
```diff
@@ -638,7 +638,7 @@ private void attemptToComplete() {
 
         // If we made it all the way through the above loop, we completed!
         // Lets flip our flag to true.
-        logger.info("Looks like all partitions are complete!  Lets wrap this up.");
+        logger.info("Looks like all partitions are resolve!  Lets wrap this up.");
```
Crim: (wat) refactor did a string replacement on this one?

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
```diff
@@ -208,6 +208,10 @@ public synchronized boolean hasVirtualSpout(final VirtualSpoutIdentifier virtual
         return runningSpouts.containsKey(virtualSpoutIdentifier);
     }
 
+    public synchronized DelegateSpout getVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifier) {
```
Crim: j docs

Crim:
I know you have a doc somewhere, but it may be helpful to updatethe PR description with clear definitions of all the states and what they mean.

stanlemon:
This PR will need a README overhaul, I'd like to work on that in a separate PR.

src/main/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinator.java
```diff
@@ -208,6 +208,19 @@ public synchronized boolean hasVirtualSpout(final VirtualSpoutIdentifier virtual
...
+     * Get a {@link DelegateSpout} instance from the {@link SpoutCoordinator}.
+     *
+     * This is useful is you want to manipulate the filter chain or alter the ending state after a {@link DelegateSpout} has
+     * been added to the {@link SpoutCoordinator}.
+     *
+     * @param virtualSpoutIdentifier identifier for teh {@link DelegateSpout} instance to get from the {@link SpoutCoordinator}.
+     * @return {@link DelegateSpout} instance
+     */
+    public synchronized DelegateSpout getVirtualSpout(final VirtualSpoutIdentifier virtualSpoutIdentifier) {
+        return runningSpouts.get(virtualSpoutIdentifier).getSpoutRunner().getSpout();
```
Crim: Probably need to handle the case where you ask for a non-existing virtual spout.  Right now I'm guessing it would NPE.

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
```diff
@@ -344,15 +339,18 @@ public void onSpoutClose(final DynamicSpout spout) {
...
     /**
-     * Starts a sideline request.
-     * @param sidelineRequest representation of the request that is being started.
+     * Start processing on the firehose for a future sideline.
+     *
+     * @param sidelineRequest sideline request.
      */
-    public void startSidelining(SidelineRequest sidelineRequest) {
+    @Override
+    public void start(SidelineRequest sidelineRequest) {
```
Crim: does there need to be any validation around, what if you attempt to START a sideline thats already in RESUME, etc..  Basically validation that this state machine can't run backwards.

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
```diff
@@ -466,11 +453,81 @@ public void stopSidelining(SidelineRequest sidelineRequest) {
...
+
+    /**
+     * Completes processing of messages that have been resumed.
+     * @param sidelineRequest sideline request.
+     */
+    @Override
+    public void resolve(SidelineRequest sidelineRequest) {
+        final SidelineRequestIdentifier id = sidelineRequest.id;
+
+        logger.info("Received FINISH sideline request");
```
Crim: comment seems wrong, FiNISHED != resolve?

src/main/java/com/salesforce/storm/spout/sideline/trigger/SidelineType.java
```diff
@@ -31,5 +31,6 @@
 public enum SidelineType {
 
     START,
```
Crim: I'd add some quick descriptions here for what each of these mean

stanlemon:
I want to add the validation mentioned by StevieP but I'd like to do it in a separate PR as part of the work on #75 as I think it'll be easier to address when that interface allows saving partitions together.

PR #82, opened 2017-11-27
first pass at creating a concrete SpoutConfig class, lots of refactor…

Crim:
…ing/cleanup left to do

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/config/AbstractConfig.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/config/AbstractConfig.java
@@ -1,0 +1,121 @@
+package com.salesforce.storm.spout.dynamic.config;
+
+import org.apache.kafka.common.config.ConfigException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Abstract/Common configuration definitions.
+ */
+public class AbstractConfig {
+    private static final Logger logger = LoggerFactory.getLogger(AbstractConfig.class);
+
+    /**
+     * Defines the Configuration.
+     */
+    private final ConfigDefinition configDefinition;
+
+    /**
+     * Stores configured values.
+     */
+    private final Map<String, Object> values = new HashMap<>();
+
+    public AbstractConfig(final ConfigDefinition configDefinition, final Map<String, Object> values) {
+        this.configDefinition = configDefinition;
+        this.values.putAll(values);
+    }
+
+    public Object get(final String name) {
+        // If we don't have a value set for this key
+        if (!values.containsKey(name)) {
+            // Return default value, or null
+            configDefinition.getDefaultValue(name);
+        }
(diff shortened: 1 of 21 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/config/AbstractConfig.java
```diff
@@ -0,0 +1,121 @@
...
+//     *
+//     * @param name Name of setting to unset.
+//     * @return Config instance.
+//     */
+//    public AbstractConfig unset(final String name) {
+//        values.remove(name);
+//        return this;
+//    }
+
+    // TODO Temp method - remove this after its no longer needed.
```
Crim: This is just here to ease refactoring in pieces.  Its expected to be removed.

src/main/java/com/salesforce/storm/spout/dynamic/config/ConfigDefinition.java
```diff
@@ -0,0 +1,126 @@
...
+            this.description = description;
+            this.category = category;
+        }
+
+        public boolean hasDefault() {
+            return this.defaultValue != null;
+        }
+    }
+
+    public String toHtmlTable() {
```
Crim: Just a stub for now, probably make this generate our readme, or we can move that to some other class.  Whatever works.

src/main/java/com/salesforce/storm/spout/dynamic/kafka/Consumer.java
```diff
@@ -771,16 +772,4 @@ public void removeConsumerState() {
         // Return TopicPartitions for our assigned partitions
         return topicPartitions;
     }
-
-    /**
-     * Utility method to reduce duplication around generating metric keyspaces for
```
Crim: This was just dead code.

src/main/java/com/salesforce/storm/spout/dynamic/FactoryManager.java
```diff
@@ -52,23 +52,23 @@
...
+    private final SpoutConfig spoutConfig;
 
     /**
      * Constructor.
      * @param spoutConfig Spout config.
      */
-    public FactoryManager(Map<String, Object> spoutConfig) {
-        // Create immutable copy of configuration.
-        this.spoutConfig = Tools.immutableCopy(spoutConfig);
+    public FactoryManager(final SpoutConfig spoutConfig) {
```
stanlemon: I like this line.

stanlemon:
I'm not a big fan of having the docs so far removed from the config declaratives in their respective classes. This seems like a step back to me. I am a proponent of a concrete class and moving in that general direction though.

Help me understand the reason for splitting SpoutConfig and DynamicConfig, is the former just supposed to be a base class that can be used elsewhere?

Crim:
So right now "SpoutConfig" is basically just a blank adhoc config object, allowing you to set a config definition, and get values out of it.  I'd like to figure out a good way to satisfy bullet 3 at the top.

"Needs to allow user to define adhoc settings for any integrations."

stanlemon:
@Crim what's the plan here? We're in merge conflict hell atm. Do we want to pursue this or abandon it for the time being?

PR #85, opened 2017-12-04
Fix ordering bug in DefaultRetryManager.

Crim:
Previously the implementation would return messageIds that were closest to the current timestamp instead of the messageIds which failed earlier.  This means if there was a high rate of sustained failures, the earlier failed messageIds may never have gotten replayed.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManager.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManager.java
@@ -184,10 +184,11 @@ public class DefaultRetryManager implements RetryManager {
         }
 
         // Populate the values
-        final Long lowestTimestampKey = entry.getKey();
+        final long lowestTimestampKey = entry.getKey();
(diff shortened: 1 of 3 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManager.java
```diff
@@ -184,10 +184,11 @@ public MessageId nextFailedMessageToRetry() {
         }
 
         // Populate the values
-        final Long lowestTimestampKey = entry.getKey();
+        final long lowestTimestampKey = entry.getKey();
```
stanlemon: This can generate a NPE can’t it?

Crim: Most map implementations disallow null keys.  And we set the key values in the map to the result of Clock's current timestamp, so that would have to return null.

I'd say its a highly unlikely scenario.

PR #86, opened 2017-12-05
Check for nulls and log accordingly

stanlemon:
Add checks and skip in loop if this should happen. This most likely occurs when there's bad serialization occurring on sideline requests.


stanlemon:
@Crim due to some prod issues I am going to merge this logging change and cut a release for 0.9.1.

Crim:
On a somewhat related note, can we push a snapshot of 0.10.x up?  I'd like to get a branch of our test "echo" topology running against it and start checking what changes are required and ensure the changelog documents it correctly

stanlemon:
@Crim Snapshot pushed earlier today.

stanlemon:
I am going to rebase this directly onto master. There's probably a better way of orchestrating commits between the two branches, but I'm not certain the best approach right now.

PR #87, opened 2017-12-06
More fair partition distribution with weird number of partitions

stanlemon:
The current partition distribution implementation attempts to distribute the maximum number of partitions to each consumer, such that when the number of consumer instances is not a multiplier of the partition count the ending instances are left with no partitions.  For example consider 4 partitions across 3 instances, the current implementation will give instances 1 & 2 two partitions each and leave instance 3 with no partitions.

The new implementation attempts to make sure every instance has at least one partition, unless there are more instances than partitions, and will distribute up to the max partition per instance number to as many instances as it can.  The trick here is that we still want the partitions ordered such that with 4 partitions across 3 instances, instance 1 should have partitions 0 and 1, and instance 2 should have partition 2 and instance 3 partition 4.  The complexity here is easier to see when you go up to 20 instances and 35 partitions as an example.  The current implementation would leave the last two instances empty, when it reality we want them each to have one.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/consumer/PartitionDistributor.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/consumer/PartitionDistributor.java
@@ -94,13 +94,13 @@ public class PartitionDistributor {
             }
 
             // How many consumer instances remain?
-            final int reaminingConsumerSlots = totalConsumers - consumerInstance;
+            final int remainingConsumerSlots = totalConsumers - consumerInstance;
 
             // We we have more slots than remaining partitions we advance to the next consumer.
             // This usually means that we have a weird number of partitions to balance across instances, like four partitions
             // across three instances, which means that last two instances will have only one partition, though the maximum
             // could be two instances. If this part confuses you, check out the test for this class.
-            if (reaminingConsumerSlots > remainingPartitions) {
+            if (remainingConsumerSlots > remainingPartitions) {
                 consumerInstance++;
             }
         }
--- a/src/main/java/com/salesforce/storm/spout/dynamic/consumer/PartitionDistributor.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/consumer/PartitionDistributor.java
@@ -26,9 +26,11 @@
 package com.salesforce.storm.spout.dynamic.consumer;
 
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/salesforce/storm/spout/dynamic/consumer/PartitionDistributorTest.java
```diff
@@ -84,6 +91,12 @@ public void testCalculatePartitionAssignment(int totalConsumers, int consumerInd
             { 1, 0, new int[]{ 0, 1, 2 }, new int[]{ 0, 1, 2 } },
             // One instance, first instance, three partitions not in order, all partitions
             { 1, 0, new int[]{ 2, 0, 1 }, new int[]{ 0, 1, 2 } },
+            // Three instances, first instance, four partitions, first two partitions
+            { 3, 0, new int[]{ 0, 1, 2, 3 }, new int[]{ 0, 1 } },
+            // Three instances, second instance, four partitions, third partition
+            { 3, 1, new int[]{ 0, 1, 2, 3 }, new int[]{ 2 } },
```
stanlemon: @c-mimms This should cover the 20/18/35 scenario, and the next one the 20/19/35 scenario.

stanlemon:
This will need to be included in 0.9.x

stanlemon:
@crim can you review this please?

Crim:
I thought mim's had it covered.  I don't have the back story to this, or understand whats changing.  Tests seem to pass so probably good?

src/main/java/com/salesforce/storm/spout/dynamic/consumer/PartitionDistributor.java
```diff
@@ -62,25 +64,48 @@
...
+            // If we've reached our maximum number of partitions per instance, advance to the next instance
+            // This means that an instance gets partitions in order, eg. [0, 1] and [2, 3], etc.
+            if (partitionByConsumer.get(consumerInstance).size() >= maxPartitionsPerInstance) {
+                consumerInstance++;
+            }
 
-        // Make a new array of integers for the partition ids
-        List<Integer> partitionIds = Lists.newArrayList();
+            // How many consumer instances remain?
+            final int reaminingConsumerSlots = totalConsumers - consumerInstance;
```
Crim: spell check here.

stanlemon:
Note that after merging I will rebase the squashed commit directly onto the 0.9 branch.

PR #88, opened 2017-12-07
Check for nulls and log accordingly, skip as appropriately

stanlemon:


stanlemon:
@c-mimms Zong-wen reported it over in the cold DC. I can definitely see it happening with stale data if a deploy happens with a change to a class that's serialized. Better safe than sorry I figure.

stanlemon:
Cherry picked this onto master as well @Crim

PR #89, opened 2017-12-08
Fix failed message leak.  the message was never removed from trackedMap in certain scenarios

Crim:
**Summary**

When VirtualSpout decides that a fail()'d message should NOT be retried, it notifies the retry manager to 'ack' it, and notifies the consumer to 'ack' it.  Unfortunately what it did NOT do was remove it from the `trackedMessages` map.  This map contains a mapping of all messages in flight, mapping their messageId => message.

Because we were not removing from this map, if a VirtualSpout had a large number of failed messages that exceeded the fail threshold, those map entries would linger around and never be cleaned up, until the vspout was closed.  This results in a memory leak.

**Solution**
I think the proper solution is to just call ack() on the messageId in this scenario.  This should mark the message as being ack'd in all the required ways.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
@@ -399,12 +399,9 @@ public class VirtualSpout implements DelegateSpout {
         if (!retryManager.retryFurther(messageId)) {
             logger.warn("Not retrying failed msgId any further {}", messageId);
 
-            // Mark it as acked in retryManager
-            retryManager.acked(messageId);
-
-            // Ack it in the consumer
-            consumer.commitOffset(messageId.getNamespace(), messageId.getPartition(), messageId.getOffset());
-
+            // Call ack to ensure it gets cleaned up
+            ack(messageId);
+            
             // Update metric
             getMetricsRecorder()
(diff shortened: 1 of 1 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/VirtualSpout.java
```diff
@@ -399,12 +399,9 @@ public void fail(Object msgId) {
         if (!retryManager.retryFurther(messageId)) {
             logger.warn("Not retrying failed msgId any further {}", messageId);
 
-            // Mark it as acked in retryManager
-            retryManager.acked(messageId);
-
-            // Ack it in the consumer
-            consumer.commitOffset(messageId.getNamespace(), messageId.getPartition(), messageId.getOffset());
-
```
Crim: Essentially these lines were missing from here, but are called in ack()

```java
 // Remove this tuple from the spout where we track things in-case the tuple fails.
trackedMessages.remove(messageId);
```

stanlemon:
Seems fine, is there a test case for this that would have failed before?

Crim:
Unfortunately not, unless we test the internals of the class -- which we do tons of other places, so on the fence about it.  What do you think should I work out a test case for it?

PR #90, opened 2017-12-08
First pass at a permanently failed output stream from the spout.

Crim:
**Use Case**
We'd like the ability to handle tuples that are considered "permanently failed".  We're defining "permanently failed" meaning that the topology attempted to process the tuple at least once and the `RetryManager` implementation has determined that the tuple should never be retried.  When this occurs, the tuple will be emitted un-anchored out a "failed" stream.  Bolts within the topology can subscribe to this "failed" stream and do its own error handling.

It should be noted (again) that tuples emitted on the "failed" stream are un-anchored and should be considered unreliable.  IE. The spout will not track it or be notified of failed or acked tuples emitted down this "failed" stream.

**Notes**

One idea was to move the RetryFailedMessageManager out of the VirtualSpouts and only have a single instance living within DynamicSpout (instead of one per VSpout).  Doing so would simplify the logic around being able to easily identify and emit out permanently failed tuples.  The trade off is this alters how we enforce fairness across VirtualSpouts.  If we continue to give priority to failed tuples being replayed over new tuples means that a single VSpout emitting tons of tuples that end up failing could overwhelm the spout's output.

Instead I added an additional flag on the Message object that allows you to mark it as permanently failed, and DynamicSpout uses that flag to determine which stream to emit down.  Not entirely the most elegant solution, so open to other ideas.

Thoughts?

**Todo**
Update changelog & readme

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
@@ -116,7 +116,7 @@ public class DynamicSpout extends BaseRichSpout {
      * Determines which output stream to emit permanently failed tuples out.
      * Gets set during open().
      */
-    private String failedOutputStreamId = null;
+    private String permanentlyFailedOutputStreamId = null;
 
     /**
      * Whether or not the spout has been previously opened.
@@ -230,11 +230,12 @@ public class DynamicSpout extends BaseRichSpout {
             return;
         }
 
-        // If this is a failed message
-        if (message.isHasFailed()) {
+        // If this is a permanently failed message.
+        if (message.isPermanentlyFailed()) {
             // Emit tuple via the output collector down the failed stream.
-            // Do not attach a messageId because this should be untracked.
-            getOutputCollector().emit(getFailedOutputStreamId(), message.getValues());
+            // Do not attach a messageId because this should be untracked.  We won't listen for
+            // any acks, fails, for messages on this stream.
+            getOutputCollector().emit(getPermanentlyFailedOutputStreamId(), message.getValues());
             return;
         }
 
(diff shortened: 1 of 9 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
```diff
@@ -491,6 +511,16 @@ String getOutputStreamId() {
         return outputStreamId;
     }
 
+    /**
+     * @return The stream that tuples that have permanently failed will be emitted out.
+     */
+    String getPermanentlyFailedOutputStreamId() {
```
Crim: Assumption here is that we're just taking the standard output stream and tacking on "_failed".

Other option is to allow explicitly defining this value.

Or we could just hard code it to something like "permanently_failed"

Or some other option?

src/test/java/com/salesforce/storm/spout/dynamic/VirtualSpoutTest.java
```diff
@@ -1656,9 +1764,7 @@ public void testSetEndingState() {
         defaultConfig.put(KafkaConsumerConfig.DESERIALIZER_CLASS, Utf8StringDeserializer.class.getName());
 
         // DynamicSpout config items
-        defaultConfig.put(SpoutConfig.PERSISTENCE_ZK_ROOT, "/sideline-spout-test");
```
Crim: Removing the ZK dependency (not needed in this test) plus calling close makes this test run in less than 1 sec, compared to 1 min 30 sec+ previously. 🎈

src/main/java/com/salesforce/storm/spout/dynamic/DynamicSpout.java
stanlemon: This seems like it should probably be driven by a configuration value with a decent default.

Crim: added SpoutConfig.PERMANENTLY_FAILED_OUTPUT_STREAM_ID => "spout.permanently_failed_output_stream_id" with a default value of "failed"

PR #91, opened 2017-12-13
Minor tweaks relating to our Metric refactor

Crim:
Going through internal projects being built against 0.10-SNAPSHOT build showed a few things missing from the Metrics refactor.  This PR attempts to close the loop on those.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/DefaultVirtualSpoutIdentifier.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/DefaultVirtualSpoutIdentifier.java
@@ -31,7 +31,7 @@ import com.google.common.base.Strings;
 /**
  * Identifier for a virtual spout.
  */
-public class DefaultVirtualSpoutIdentifier implements VirtualSpoutIdentifier {
+public final class DefaultVirtualSpoutIdentifier implements VirtualSpoutIdentifier {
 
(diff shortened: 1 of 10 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/DefaultVirtualSpoutIdentifier.java
```diff
@@ -81,6 +81,6 @@ public boolean equals(Object obj) {
      */
     @Override
     public int hashCode() {
-        return id != null ? id.hashCode() : 0;
```
Crim: id cannot be null.

src/main/java/com/salesforce/storm/spout/dynamic/MessageId.java
```diff
@@ -93,11 +101,14 @@ public boolean equals(Object other) {
 
     @Override
     public int hashCode() {
-        int result = getNamespace().hashCode();
-        result = 31 * result + getPartition();
-        result = 31 * result + (int) (getOffset() ^ (getOffset() >>> 32));
-        result = 31 * result + getSrcVirtualSpoutId().hashCode();
-        return result;
+        if (hash == 0) {
```
Crim: Micro optimization since these are often referred to in maps.

PR #95, opened 2017-12-14
Remove references to sideline from DynamicSpout tests

Crim:
Continue to work towards removing Sideline dependency out of DynamicSpout core package space.

Removes dependencies within tests to Sideline classes, but the tests themselves don't actually depend on any sideline logic.  So swapped them out with their Default implementations.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
+++ b/src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
@@ -429,16 +429,11 @@ public class SidelineSpoutHandler implements SpoutHandler, SidelineController {
             );
 
             if (sidelinePayload == null) {
-                logger.warn("Sideline {} on partition {} payload was null, this is probably a serialization problem.");
+                logger.error("Attempting to load sideline {} for partition {} and it was null.", id, consumerPartition.partition());
                 continue;
             }
 
             logger.info("Loaded sideline payload for {} = {}", consumerPartition, sidelinePayload);
-
-            if (sidelinePayload == null) {
-                logger.error("Attempting to load sideline {} for partition {} and it was null.", id, consumerPartition.partition());
-                continue;
-            }
 
             if (sidelinePayload.startingOffset == null) {
                 logger.error("Loaded sideline {} for partition {} but the starting offset was null", id, consumerPartition.partition());
--- a/src/main/java/com/salesforce/storm/spout/sideline/persistence/ZookeeperPersistenceAdapter.java
+++ b/src/main/java/com/salesforce/storm/spout/sideline/persistence/ZookeeperPersistenceAdapter.java
@@ -242,7 +242,7 @@ public class ZookeeperPersistenceAdapter implements PersistenceAdapter {
 
                 for (final String partition : partitions) {
                     consumerPartitions.add(
-                        new ConsumerPartition(namespace, Integer.valueOf(partition))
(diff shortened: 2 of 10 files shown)


src/test/java/com/salesforce/storm/spout/dynamic/VirtualSpoutTest.java
```diff
@@ -494,7 +476,7 @@ public void testNextTupleReturnsNullWhenFiltered() {
             null,
             null
         );
-        virtualSpout.getFilterChain().addStep(new SidelineRequestIdentifier("Foobar"), filterStep);
+        virtualSpout.getFilterChain().addStep(new DefaultFilterChainStepIdentifier("Foobar"), filterStep);
```
Crim: Doesnt look like anything in this test specifically depends on SidelineRequestIdentifier.  Swapped it with the default filter chain step id.

src/test/java/com/salesforce/storm/spout/dynamic/VirtualSpoutTest.java
```diff
@@ -1507,54 +1460,6 @@ public void testCloseWithCompletedFlagSetToTrue() throws NoSuchFieldException, I
         verify(mockConsumer, never()).flushConsumerState();
     }
 
-    /**
-     * Test calling close, verifies what happens if the completed flag is true.
-     * Verifies what happens if SidelineRequestIdentifier is null.
-     */
-    @Test
-    public void testCloseWithCompletedFlagSetToTrueNoSidelineREquestIdentifier() throws NoSuchFieldException, IllegalAccessException {
```
Crim: This test is a duplicate of the test directly above it.  I believe its a hold over from when we had sideline logic more closely intertwined into the VirtualSpout class.

src/main/java/com/salesforce/storm/spout/sideline/handler/SidelineSpoutHandler.java
```diff
@@ -429,17 +429,12 @@ public void stopSidelining(SidelineRequest sidelineRequest) {
             );
```
Crim: This is just duplicated code.  We were checking to see if sidelinePayload was equal to null twice.  Didn't make sense, because it can't be null after we check it the first time.  I just replaced the top check with the bottom check.  Only real difference is its now an error instead of warn, and logs more details.

src/main/java/com/salesforce/storm/spout/sideline/persistence/ZookeeperPersistenceAdapter.java
```diff
@@ -242,7 +242,7 @@ public void clearSidelineRequest(final SidelineRequestIdentifier id, final Consu
 
                 for (final String partition : partitions) {
                     consumerPartitions.add(
-                        new ConsumerPartition(namespace, Integer.valueOf(partition))
```
Crim: Logically the same, Integer.valueOf() returns an Integer where Integer.parseInt() returns an int, which is what ConsumerPartition's constructor takes.

src/main/java/com/salesforce/storm/spout/sideline/trigger/example/ZookeeperWatchTrigger.java
```diff
@@ -165,7 +166,7 @@ public void open(final Map config) {
                     curator
                         .create()
                         .creatingParentsIfNeeded()
-                        .forPath(root, "".getBytes());
```
Crim: Logically the same, w/o having to create a new String object first.

PR #96, opened 2017-12-14
Switching test to use the in memory persistence adapters

stanlemon:
This is an attempt to decouple the `SidelineSpoutTest` from Zookeeper.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/persistence/InMemoryPersistenceAdapter.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/persistence/InMemoryPersistenceAdapter.java
@@ -37,13 +37,13 @@ public class InMemoryPersistenceAdapter implements PersistenceAdapter {
     /**
      * In memory store for this adapter's data.
      */
-    public static Map<String, Long> storedConsumerState = Maps.newHashMap();
+    public static Map<String, Long> storedConsumerState = Maps.newConcurrentMap();
 
     /**
      * Reset the internal memory store.
      */
     public static void reset() {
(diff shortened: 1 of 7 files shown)


stanlemon:
While digging into this I also tried swapping in the `MockConsumer` and that proved to be much more hairy than I thought it would be. This test is tightly coupled to KafkaRecord instances and it's just going to take some serious time to decouple them and maintain the test.

src/main/java/com/salesforce/storm/spout/dynamic/persistence/InMemoryPersistenceAdapter.java
```diff
@@ -35,27 +35,23 @@
 public class InMemoryPersistenceAdapter implements PersistenceAdapter {
 
     /**
-     * Used within tests to more easily validate assertions.
+     * In memory store for this adapter's data.
      */
-    public static PersistConsumerStateCallback<String, Integer, Long> persistConsumerStateCallback = (consumerId, partitionId, offset) -> {
-        // No-op by default.
-    };
+    public static Map<String, Long> storedConsumerState = Maps.newHashMap();
```
Crim: This feels like it'll break as soon as you fire up more than one vspout will it not?  Being static and shared across all instances?

stanlemon: Probably just need to make this a concurrent map in both adapters to be safe.

stanlemon:
Based upon your comments @Crim I am going to close this for now.

PR #97, opened 2018-01-17
Set sideline defaults when loading config

stanlemon:
Ensures that defaults are properly applied to the config when it is passed into the sidelining code.


stanlemon:
@Crim I think this an oversight when we split the configuration up.

PR #98, opened 2018-01-18
Allow null keys in the Utf8Deserializer

stanlemon:


src/main/java/com/salesforce/storm/spout/dynamic/kafka/deserializer/Utf8StringDeserializer.java
```diff
@@ -36,7 +36,8 @@
     @Override
     public Values deserialize(String topic, int partition, long offset, byte[] key, byte[] value) {
         return new Values(
-            new String(key, Charsets.UTF_8),
```
Crim: makes sense that they could also have null values as well.  We should probably handle both.  I'd argue that if value is null, just return null from the deserializer, which I think in turn skips the message.

stanlemon: Updated to account for that, check it once more please @Crim

Crim:
Maybe update the CHANGELOG, looks good :shipit:

PR #104, opened 2018-02-01
Swap around code in try block

Crim:


Crim:
I think it was just super minor code cleanup.  Basically just wrapping the try/catch block around the part that could actually throw the exception, and narrowing down the caught exception to the specific type that could be thrown.

Still seems valid to me.

Crim:
Merge if you agree or close er out.

stanlemon:
Yup, my memory is coming back to me, I think we swallowed something on start in production that should have bubbled all the way down.  Going to go ahead and merge this.

PR #110, opened 2018-06-25
Change the default retry delay to have exponential growth

mmoldavan:
While debugging retries, I noticed that the current retry behavior follows linear growth (2, 4, 6, 8 seconds). However, the comments on the DefaultRetryManager class and retry variables imply expontential growth (2, 4, 8, 16, etc.).

This fixes the retry code to grow exponentially as expected.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManager.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManager.java
@@ -121,7 +121,7 @@ public class DefaultRetryManager implements RetryManager {
 
         // Determine when we should retry this msg next
         // Calculate how many milliseconds to wait until the next retry
-        long additionalTime = (long) (getInitialRetryDelayMs() * Math.pow(getRetryDelayMultiplier(), failCount));
+        long additionalTime = (long) (getInitialRetryDelayMs() * Math.pow(getRetryDelayMultiplier(), failCount - 1));
         if (additionalTime > getRetryDelayMaxMs()) {
             // If its over our configured max delay, use max delay
             additionalTime = getRetryDelayMaxMs();
--- a/src/main/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManager.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManager.java
@@ -121,7 +121,7 @@ public class DefaultRetryManager implements RetryManager {
 
         // Determine when we should retry this msg next
         // Calculate how many milliseconds to wait until the next retry
-        long additionalTime = (long) (failCount * getInitialRetryDelayMs() * getRetryDelayMultiplier());
+        long additionalTime = (long) (getInitialRetryDelayMs() * Math.pow(getRetryDelayMultiplier(), failCount));
         if (additionalTime > getRetryDelayMaxMs()) {
             // If its over our configured max delay, use max delay
             additionalTime = getRetryDelayMaxMs();
--- a/src/test/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManagerTest.java
+++ b/src/test/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManagerTest.java
@@ -35,6 +35,7 @@ import org.junit.Test;
 import java.time.Clock;
 import java.time.Instant;
 import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.Map;
 import java.util.Queue;
 
@@ -541,7 +542,7 @@ public class DefaultRetryManagerTest {
         retryManager.setClock(
             Clock.fixed(
             Instant.ofEpochMilli(FIXED_TIME + (long) (expectedMinRetryTimeMs * Math.pow(expectedDelayMultiplier, 0))),
-            ZoneId.of("UTC")
+            ZoneOffset.UTC
         ));
 
         // Now messageId1 should expire next,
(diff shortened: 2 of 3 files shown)


src/main/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManager.java
```diff
@@ -121,7 +121,7 @@ public void failed(MessageId messageId) {
 
         // Determine when we should retry this msg next
         // Calculate how many milliseconds to wait until the next retry
-        long additionalTime = (long) (failCount * getInitialRetryDelayMs() * getRetryDelayMultiplier());
+        long additionalTime = (long) (getInitialRetryDelayMs() * Math.pow(getRetryDelayMultiplier(), failCount));
```
noshirpatel-salesforce: We should remove the + 1 when we initialize failCount so that it starts at 0.  Otherwise we never use the initial delay, but skip directly to what should be the delay for the second retry.

mmoldavan: It makes more sense to me to leave the failCount initialization as is. Otherwise, it would be off by one as far as the true number of times a message has failed, and there is other logic scattered about based on it. Instead, we should just subtract 1 here when calculating the delay.

src/test/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManagerTest.java
```diff
@@ -539,8 +539,10 @@ public void testNextFailedMessageToRetryWithExpiring() {
 
         // Now advance time by exactly expectedMinRetryTimeMs milliseconds
         retryManager.setClock(
-            Clock.fixed(Instant.ofEpochMilli(FIXED_TIME + (long) (expectedMinRetryTimeMs * expectedDelayMultiplier)), ZoneId.of("UTC"))
-        );
+            Clock.fixed(
+            Instant.ofEpochMilli(FIXED_TIME + (long) (expectedMinRetryTimeMs * Math.pow(expectedDelayMultiplier, 0))),
```
stanlemon: This line and the next one should be indented.

src/test/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManagerTest.java
```diff
@@ -558,7 +560,7 @@ public void testNextFailedMessageToRetryWithExpiring() {
         assertNull("Should be null", retryManager.nextFailedMessageToRetry());
 
         // Advance time again, by 2x expected retry time, plus a few MS
```
stanlemon: This comment needs to be updated doesn’t it?

src/test/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManagerTest.java
```diff
@@ -575,7 +577,7 @@ public void testNextFailedMessageToRetryWithExpiring() {
         validateTupleNotInFailedSetButIsInFlight(retryManager, messageId2);
 
         // Calculate time for 3rd fail, against new fixed time
```
stanlemon: This one as well.

src/test/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManagerTest.java
```diff
@@ -539,8 +539,10 @@ public void testNextFailedMessageToRetryWithExpiring() {
 
         // Now advance time by exactly expectedMinRetryTimeMs milliseconds
         retryManager.setClock(
-            Clock.fixed(Instant.ofEpochMilli(FIXED_TIME + (long) (expectedMinRetryTimeMs * expectedDelayMultiplier)), ZoneId.of("UTC"))
-        );
+            Clock.fixed(
+            Instant.ofEpochMilli(FIXED_TIME + (long) (expectedMinRetryTimeMs * Math.pow(expectedDelayMultiplier, 0))),
+            ZoneId.of("UTC")
```
stanlemon: I know the old test did this as well, but fwiw I think you can do `ZoneOffset.UTC`

src/main/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManager.java
stanlemon: I agree with @mmoldavan here though I totally get where you’re coming from @noshirpatel-salesforce . I’d prefer we leave it as Matt suggests.

One thing I noticed though, for some reason we don’t supply the 1 in getOrDefault() here https://github.com/mmoldavan/storm-dynamic-spout/blob/4c6081fe2570f1576b3f15e47878d8326e6e0b7d/src/main/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManager.java#L119 which seems strange. This line also could stand to ha e a comment stating basically what you said above @mmoldavan .

stanlemon:
@Crim side note to the PR, I kind of think we should rename this class in 0.10 to ExponentialBackoffRetryManager, what do you think of that?

stanlemon:
@mmoldavan this looks great, thanks for the contribution. I left a few comments, nothing major. Once we get those covered can you add an item to the change log and we’ll get this merged in. Do you need me to backport this over to 0.9?

src/test/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManagerTest.java
mmoldavan: the checkstyles didn't like the indention for some reason.

mmoldavan:
👍 I will fix up the comments and changelog. If stardust can be bumped to 0.10 easily, I don't think it needs backporting necessarily.

src/main/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManager.java
noshirpatel-salesforce: Yeah, I see.  That is the increment.  Works for me.

src/test/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManagerTest.java
stanlemon: That's odd! Roger that though, thanks for the explanation.

stanlemon:
@mmoldavan 0.10 is still in development and none of the tooling internal for Stardust supports it yet so I'll make sure to back port this to 0.9. I don't imagine that being much of a problem.  Once that's done I'll cut new releases.

stanlemon:
Thanks for the contribution @mmoldavan!

I'll work on back porting this next week, feel free to ping me for updates.

PR #111, opened 2018-07-05
Update dependencies and cleanup

stanlemon:
This pull request does:
- Updates various dependencies in pom.xml
- Updates junit and kafka-junit to the v5 versions
- Fixed a bunch of checkstyle warnings around docs
- Removed asterisks from license header (IDE friendly)
- Updates copyrights
- Address some unchecked handling and improves typing on certain config objects

This pull request does NOT:
- Cleanup all of the Mockito matchers
- Resolve deprecated code issues, such as those with storm metrics
- Resolve unchecked handling in tests
- Updated curator dependencies (requires further testing)

I will open a followup PR with some of the _does not_ items, this one has gotten too big already.


pom.xml
```diff
@@ -125,18 +120,18 @@
...
+            <artifactId>commons-collections4</artifactId>
+            <version>4.1</version>
         </dependency>
 
-        <!-- Curator is not compatible with Guava > 20.0 :( -->
+        <!-- Google's Guava library, makes life easier through common tooling -->
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>20.0</version>
```
Crim: If I recall this used to break Curator?  Is this no longer the case?

pom.xml
```diff
@@ -328,48 +362,22 @@
...
+                    </dependency>
+                    <dependency>
+                        <groupId>org.junit.jupiter</groupId>
+                        <artifactId>junit-jupiter-engine</artifactId>
+                        <version>${junit5Version}</version>
                     </dependency>
                 </dependencies>
                 <configuration>
-                    <!-- ${argLine} gets defined by jacoco plugin -->
-                    <argLine>${argLine} -Xmx1g</argLine>
```
Crim: Do we no longer need the extra heap size?

pom.xml
```diff
@@ -474,7 +482,7 @@
                         <dependency>
                             <groupId>com.puppycrawl.tools</groupId>
                             <artifactId>checkstyle</artifactId>
-                            <version>8.0</version>
+                            <version>8.11</version>
```
Crim: We may want to copy over the latest checkstyle ruleset xml file?  Or maybe you did already haven't read that far into this PR yet.

src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorHelper.java
```diff
@@ -97,11 +96,9 @@ public void writeJson(final String path, final Object data) {
      * will deserialize the JSON to a concrete class.
      *
      * @param path node containing JSON to read from.
-     * @param <K> key to the json field.
-     * @param <V> value of the json field.
      * @return map representing the JSON stored within the zookeeper node.
      */
-    public <K, V> Map<K, V> readJson(final String path) {
+    public HashMap readJson(final String path) {
```
Crim: Why change this?  Seems like its moving backwards?

src/main/java/com/salesforce/storm/spout/dynamic/retry/DefaultRetryManager.java
```diff
@@ -133,7 +133,7 @@ public void failed(MessageId messageId) {
         if (failCount > 1) {
             // Make sure they're removed.  This kind of sucks.
             // This may not be needed in reality...just because of how we've setup our tests :/
-            for (Queue queue: failedMessageIds.values()) {
+            for (Queue queue : failedMessageIds.values()) {
```
Crim: Only because you changed the line, can this be marked final?

src/main/java/com/salesforce/storm/spout/sideline/config/DocTask.java
```diff
@@ -63,7 +63,7 @@ private static void generateMetricDocs(final Path inputPath) throws IOException
     private static void generateConfigDocs(final Path inputPath) throws IOException {
         final String tagArg = "SIDELINE_CONFIGURATION";
         final List<ClassSpec> classSpecs = new ArrayList<>();
-        classSpecs.add(new ClassSpec(SidelineConfig.class, SidelineConfig.setDefaults(new HashMap())));
+        classSpecs.add(new ClassSpec(SidelineConfig.class, SidelineConfig.setDefaults(Maps.newHashMap())));
```
Crim: and here I was working towards removing the dependency on common collections by removing these :p

src/test/java/com/salesforce/storm/spout/dynamic/DynamicSpoutTest.java
```diff
@@ -159,14 +149,14 @@ public void testCannotOpenMultipleTimes() throws InterruptedException {
...
-            spout.close();
-        }
+        final Throwable thrown = Assertions.assertThrows(IllegalStateException.class, () ->
+            spout.open(config, topologyContext, spoutOutputCollector)
+        );
+
+        MatcherAssert.assertThat(thrown.getMessage(), Matchers.containsString("opened"));
+
+        // Cleanup.
+        spout.close();
```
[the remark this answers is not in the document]
Crim: I think you can just move the spout.close() up above the assertThat() call.

src/test/java/com/salesforce/storm/spout/dynamic/FactoryManagerTest.java
```diff
@@ -55,30 +52,26 @@
...
-        // Try with UTF8 String deserializer
-        final Map config = Maps.newHashMap();
-        final FactoryManager factoryManager = new FactoryManager(config);
-
-        // We expect this to throw an exception.
-        expectedExceptionCreateNewFailedMsgRetryManagerInstanceMissingConfig.expect(IllegalStateException.class);
-        factoryManager.createNewFailedMsgRetryManagerInstance();
+        Assertions.assertThrows(IllegalStateException.class, () -> {
+            // Try with UTF8 String deserializer
+            final Map config = Maps.newHashMap();
```
Crim: Probably should move this up and out of the code block.  Otherwise this could falsely trigger on the constructor instead of the createNewFialedMsgRetryManagerInstance() call.

src/test/java/com/salesforce/storm/spout/dynamic/FactoryManagerTest.java
```diff
@@ -114,21 +107,19 @@ public void testCreateNewFailedMsgRetryManager(final Class clazz) {
...
-        // We expect this to throw an exception.
-        expectedExceptionCreateNewPersistenceAdapterInstanceMissingConfig.expect(IllegalStateException.class);
-        factoryManager.createNewPersistenceAdapterInstance();
+        Assertions.assertThrows(IllegalStateException.class, () -> {
+            // Try with UTF8 String deserializer
+            final Map config = Maps.newHashMap();
+            final FactoryManager factoryManager = new FactoryManager(config);
+
+            // We expect this to throw an exception.
+            factoryManager.createNewPersistenceAdapterInstance();
```
Crim: same here.  Move everything abvoe this line out of the assertThrows() block.

src/test/java/com/salesforce/storm/spout/dynamic/FactoryManagerTest.java
```diff
@@ -158,27 +149,24 @@ public void testCreateNewPersistenceAdapterUsingDefaultImpl() {
...
-        // We expect this to throw an exception.
-        expectedExceptionCreateNewMessageBufferInstanceMissingConfig.expect(IllegalStateException.class);
-        factoryManager.createNewMessageBufferInstance();
+        Assertions.assertThrows(IllegalStateException.class, () -> {
+            // Try with UTF8 String deserializer
+            final Map config = Maps.newHashMap();
+            final FactoryManager factoryManager = new FactoryManager(config);
+
+            // We expect this to throw an exception.
+            factoryManager.createNewMessageBufferInstance();
```
Crim: samesies.

src/test/java/com/salesforce/storm/spout/dynamic/VirtualSpoutTest.java
```diff
@@ -271,14 +265,12 @@ public void testCallingOpenTwiceThrowsException() {
...
 
-        // Set expected exception
-        try {
-            expectedExceptionCallingOpenTwiceThrowsException.expect(IllegalStateException.class);
-            virtualSpout.open();
-        } finally {
-            // Ensure that we call close.
-            virtualSpout.close();
-        }
+        Assertions.assertThrows(IllegalStateException.class, () ->
```
Crim: these also worry me the same as the other ones.  You may want to verify that if  virtualSpout.open() does NOT throw an illegalstateexception, or some other class, that vspout.close() below gets called.

src/test/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinatorTest.java
```diff
@@ -768,16 +758,11 @@ public void testAddDuplicateSpout() {
...
-            spoutCoordinator.close();
-
-            // Re-throw the expected exception so the test passes.
-            throw exception;
-        }
+        Assertions.assertThrows(SpoutAlreadyExistsException.class, () ->
+            spoutCoordinator.addVirtualSpout(spout2)
+        );
+
+        spoutCoordinator.close();
```
Crim: more concerns around callign close(), same as others.

src/test/java/com/salesforce/storm/spout/dynamic/coordinator/SpoutCoordinatorTest.java
```diff
@@ -876,18 +861,14 @@ public void testAddAndRemoveVirtualSpout() {
...
-            spoutCoordinator.close();
-
-            // Rethrow expected exception.
-            throw exception;
-        }
+            spoutCoordinator.removeVirtualSpout(virtualSpoutIdentifier)
+        );
+
+        // Make sure to close out SpoutCoordinator
+        spoutCoordinator.close();
```
Crim: same as above.

Crim:
My only concerns are on the tests where <instance>.close() calls got moved out of finally blocks.  Those took me ages to track down spots where .close() wasn't being called correctly leaving zombie processes running all over.  We should just double check that even when the assertions fail, that close() gets called.

Should be simple to verify by just adjusting one of them to not throw the expected exception and see what happens.

pom.xml
stanlemon: It probably does, but I'm going to work through the curator stuff next.

pom.xml
stanlemon: Does not appear to, I haven't had issues locally or on travis. 🤞

pom.xml
stanlemon: I did, I also kept a few differences and added the unused import filter.

src/main/java/com/salesforce/storm/spout/dynamic/persistence/zookeeper/CuratorHelper.java
stanlemon: To avoid unchecked issues. The way this one works, it's going to always be a `HashMap` so the explicitness doesn't take anything away, just makes everything more specific.

stanlemon:
Good catch on those finally blocks, I think I got those all back in now and addressed the other feedback as well.

PR #112, opened 2018-07-06
Cleanup dependencies

stanlemon:


src/main/java/com/salesforce/storm/spout/dynamic/Tools.java
```diff
@@ -49,14 +47,8 @@
      * @return sshallow cloned map that is immutable.
      */
     public static <K,V> Map<K,V> immutableCopy(Map<K,V> sourceMap) {
-        // If we're already dealing with an UnmodifiableMap
-        if (sourceMap instanceof UnmodifiableMap) {
```
stanlemon: Interestingly this is NOT the `UnmodifiableMap` used by `Collections.unmodifiableMap()` so this never actually did anything.

src/main/java/com/salesforce/storm/spout/documentation/DocGenerator.java
```diff
@@ -117,7 +117,7 @@ private void injectContent(final String content) throws IOException {
         Files.copy(inputPath, tempOutPath);
 
         try (BufferedReader readmeReader = Files.newBufferedReader(inputPath, StandardCharsets.UTF_8);
-            PrintWriter readmePrintWriter = new PrintWriter(Files.newBufferedWriter(tempOutPath, StandardCharsets.UTF_8))) {
+             PrintWriter readmePrintWriter = new PrintWriter(Files.newBufferedWriter(tempOutPath, StandardCharsets.UTF_8))) {
```
Crim: extra space here?

src/main/java/com/salesforce/storm/spout/dynamic/Tools.java
Crim: should we add the correct instance check here?

src/main/java/com/salesforce/storm/spout/dynamic/metrics/LogRecorder.java
```diff
@@ -43,12 +42,11 @@
 public class LogRecorder implements MetricsRecorder {
 
     private static final Logger logger = LoggerFactory.getLogger(LogRecorder.class);
-    private final Map<String, Long> counters = Maps.newConcurrentMap();
-    private final Map<String, CircularFifoQueue> averages = Maps.newConcurrentMap();
```
Crim: Was averages just not used?

src/main/java/com/salesforce/storm/spout/dynamic/Tools.java
stanlemon: I looked into this, the problem is the actual class is private to the Collections class, so we can’t.

stanlemon: For reference http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/Collections.java#l1499

src/main/java/com/salesforce/storm/spout/dynamic/metrics/LogRecorder.java
stanlemon: Correct, which made this one easy to deal with. 😎

PR #114, opened 2018-07-09
Rename DefaultRetryManager to ExponentialRetryManager

stanlemon:
This class name is more descriptive of the behavior of the class.


stanlemon:
Heads up @mmoldavan

Crim:
Kinda dislike renaming it just to rename it because it will break backwards compat, but 🤷‍♂️

stanlemon:
I totally get that @Crim.  0.10 (unreleased) is a big BC break from 0.9, considering everything that changed around sidelining. I think the new name is a lot more clear as to what the class does. Furthermore, I would rather make these sort of changes before we get to a 1.0.

PR #115, opened 2018-07-09
Upgrade curator

stanlemon:
Upgrades to the latest curator version and follows the inclusion/exclusion pattern that storm uses.


stanlemon:
@Crim reading around I'm pretty sure that the issues we hit before were because curator wants to pull in zookeeper 3.5.  This should work around that. 🤞

Issue #117, opened 2018-07-10
Upgrade to Storm 1.2 metrics methods


stanlemon:
Storm 1.1 and lower offered a method on `TopologyContext` called `registerMetrics()` (see https://github.com/apache/storm/blob/1.1.x-branch/storm-core/src/jvm/org/apache/storm/task/TopologyContext.java).  In Storm 1.2 these methods were deprecated and `registerTimer()`, `registerHistogram()`, `registerMeter()`, `registerCounter()` and `registerGauge()` were added (see https://github.com/apache/storm/blob/v1.2.2/storm-core/src/jvm/org/apache/storm/task/TopologyContext.java#L396-L418).  Within this library we represent several metric types via the `MetricsRecorder` interface, specifically `countBy()`, `assignValue()` and `recordTimer()` along with a set of helpers for each. Our current metrics methods map to three of these, but we have gaps for meters (_measures mean throughput and one-, five-, and fifteen-minute exponentially-weighted moving average throughputs_) and histograms (calculates the distribution of a value).

In the short term (0.10) we can swap `StormRecorder` (the implementation of `MetricsRecorder` that leverages `TopologyContext` to call `registerMetrics()` to use these new methods. This does then impose a hard requirement on Storm 1.2 because the new metrics methods did not exist before that release.  Currently 0.10 requires 1.2, but it should function fine with older version - this would not be the case with a metrics change. Longer term do we need to consider exposing the other two metrics methods?  As it stands today we don't have a use case currently, but we do pass the `MetricsRecorder` instance downstream to many classes that can be overridden through third party extension.

stanlemon:
This is resolved with #119.

PR #118, opened 2018-07-12
Metrics Refactoring

stanlemon:
First part in my work on #117, this takes several bits of code that are reused between `StormRecorder` and `LogRecorder` and breaks them out so that they can be used in a forthcoming `DropwizardRecorder` implementation.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/metrics/LogRecorder.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/metrics/LogRecorder.java
@@ -25,12 +25,9 @@
 
 package com.salesforce.storm.spout.dynamic.metrics;
 
-import org.apache.storm.task.TopologyContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.slf4j.helpers.MessageFormatter;
 
-import java.time.Clock;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -48,33 +45,7 @@ public class LogRecorder implements MetricsRecorder {
     private final Map<String, Long> counters = new ConcurrentHashMap<>();
     private final Map<String, Object> assignedValues = new ConcurrentHashMap<>();
 
-    // For storing timer start values
-    private final Map<String, Long> timerStartValues = new ConcurrentHashMap<>();
-
-    @Override
-    public void open(Map spoutConfig, TopologyContext topologyContext) {
-    }
-
-    @Override
-    public void close() {
-        // Noop
-    }
-
-    @Override
-    public void count(final MetricDefinition metric) {
-        countBy(metric, 1L, new Object[0]);
-    }
-
-    @Override
-    public void count(final MetricDefinition metric, final Object... metricParameters) {
-        countBy(metric, 1L, metricParameters);
-
-    }
-
-    @Override
-    public void countBy(final MetricDefinition metric, final long incrementBy) {
-        countBy(metric, incrementBy, new Object[0]);
-    }
+    private final TimerManager timerManager = new TimerManager();
 
     @Override
     public void countBy(final MetricDefinition metric, final long incrementBy, final Object... metricParameters) {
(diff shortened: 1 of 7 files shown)


src/test/java/com/salesforce/storm/spout/dynamic/metrics/KeyBuilderTest.java
```diff
@@ -0,0 +1,97 @@
...
+            "Metric without a prefix comes out correctly",
+            metricName,
+            equalTo(
+                keyBuilder.build(metricDefinition, new Object[0])
+            )
+        );
+    }
+
+    /**
+     * Test that a {@link KeyBuilder} with parameters have them properly formatted.
```
Crim: just curious, what happens if you do

```java
final String metricName = "custom_metric.{}.{}";
....
keyBuilder.build(metricDefinition, new String[]{ param1 })
```
?

stanlemon: I think the second set of {} just hangs there. String replacement is handled by the same formatted used for logging.

src/main/java/com/salesforce/storm/spout/dynamic/metrics/LogRecorder.java
```diff
@@ -42,36 +39,13 @@
...
-
-    @Override
-    public void open(Map spoutConfig, TopologyContext topologyContext) {
-    }
 
-    @Override
-    public void close() {
-        // Noop
-    }
+    private final KeyBuilder keyBuilder = new KeyBuilder(null);
```
Crim: Should this be set in open()?  Is KeyBuilder serializable?

src/main/java/com/salesforce/storm/spout/dynamic/metrics/TimerManager.java
```diff
@@ -0,0 +1,120 @@
...
+    private Clock getClock() {
+        return this.clock;
+    }
+
+    /**
+     * Override the manager's clock instance, this should only be done in tests.
+     *
+     * @param clock clock instance for testing
+     */
+    void setClock(final Clock clock) {
```
Crim: Wonder if this should be a package protected constructor?

src/main/java/com/salesforce/storm/spout/dynamic/metrics/LogRecorder.java
stanlemon: Hmmm... good call. It probably serializes fine, but this is a good reason to put it in open(), though I hope no one uses this recorder in production.

src/main/java/com/salesforce/storm/spout/dynamic/metrics/TimerManager.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Start a timer based upon a provided key.
+     *
+     * Defaults start time to right now.
+     *
+     * @param key key for timer
+     */
+    void start(final String key) {
+        start(key, getClock().millis());
```
Crim: may wanna have null checks

src/main/java/com/salesforce/storm/spout/dynamic/metrics/TimerManager.java
stanlemon: Tried that, but it didn’t work with the test since I need to pass in a new clock on the same instance. Check out the start stop test.

src/main/java/com/salesforce/storm/spout/dynamic/metrics/TimerManager.java
stanlemon: On the clock? I’ll probably add that to the getter. Good call.

src/test/java/com/salesforce/storm/spout/dynamic/metrics/KeyBuilderTest.java
stanlemon: @Crim I went ahead and added a test to confirm my hypothesis.

src/main/java/com/salesforce/storm/spout/dynamic/metrics/TimerManager.java
stanlemon: Added some null checks for both key and clock as well as covering tests.

Crim:
seems legit

PR #119, opened 2018-07-13
Add DropwizardRecorder implementation of MetricsRecorder for use with Storm 1.2

stanlemon:
Related to #117

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/metrics/DropwizardRecorder.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/metrics/DropwizardRecorder.java
@@ -31,7 +31,7 @@ import com.codahale.metrics.Timer;
 import com.salesforce.storm.spout.dynamic.config.SpoutConfig;
 import org.apache.storm.task.TopologyContext;
 
-import java.util.HashMap;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
@@ -48,9 +48,9 @@ public class DropwizardRecorder implements MetricsRecorder {
     private KeyBuilder keyBuilder;
     private TimerManager timerManager = new TimerManager();
 
-    private final Map<MetricDefinition, Timer> timers = new HashMap<>();
-    private final Map<MetricDefinition, Counter> counters = new HashMap<>();
-    private final Map<MetricDefinition, Histogram> histograms = new HashMap<>();
+    private final Map<MetricDefinition, Timer> timers = new ConcurrentHashMap<>();
+    private final Map<MetricDefinition, Counter> counters = new ConcurrentHashMap<>();
+    private final Map<MetricDefinition, Histogram> histograms = new ConcurrentHashMap<>();
 
     @Override
     public void open(final Map<String, Object> spoutConfig, final TopologyContext topologyContext) {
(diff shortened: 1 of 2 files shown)


stanlemon:
@Crim I've got mixed feelings on tests for this guy. I can definitely write them, they'll just be mocks explicitly verifying internal calls and I'm not entirely sure that's beneficial. By breaking out the `KeyBuilder` and the `TimerManager` I feel like the actual business logic of the recorders are properly covered. What do you think?

src/main/java/com/salesforce/storm/spout/dynamic/metrics/DropwizardRecorder.java
```diff
@@ -0,0 +1,147 @@
...
+     * the {@link MultiAssignableMetric}.
+     *
+     * This particularly method should be considered experimental at this point as we may revise it before finalizing the release.
+     *
+     * @param metric metric definition.
+     * @param value value to be assigned.
+     * @param metricParameters when a {@link MetricDefinition} supports interpolation on it's key, for example "foo.{}.bar" the {}
+     */
+    @Override
+    public void assignValue(MetricDefinition metric, Object value, Object... metricParameters) {
```
Crim: So I would have thought that a assigning a value would be using a gauge under the hood?

src/main/java/com/salesforce/storm/spout/dynamic/metrics/DropwizardRecorder.java
```diff
@@ -0,0 +1,147 @@
...
+    public long stopTimer(MetricDefinition metric, Object... metricParameters) {
+        final long elapseMs = timerManager.stop(keyBuilder.build(metric, metricParameters));
+        recordTimer(metric, elapseMs, metricParameters);
+        return elapseMs;
+    }
+
+    @Override
+    public void recordTimer(MetricDefinition metric, long elapsedTimeMs, Object... metricParameters) {
+        if (!timers.containsKey(metric)) {
+            timers.put(metric, getTopologyContext().registerTimer(keyBuilder.build(metric, metricParameters)));
```
Crim: I'm not super familiar w/ the new metrics system in storm.  Under the old system you had to register all of your metrics upfront during the open() phase.  Not being able to support dynamic metrics was generally the reason why we built this abstraction in the first place.

So with that said... is it now possible to register them adhoc after the topology is up and running?

Crim: So if it IS possible to register metrics dynamically, if my memory serves me, at least for the old metric system, there was a way to ask the topologyContext for a metric instance by key.  If that's still the case, it may mean that keeping instances of the metrics locally to this class in the hashmap may no longer be required.

Crim: ooh so here's another possible issue.  With the old metric recorder, we didn't ensure thread safety into these methods, under the assumption that if two threads tried to create the same metric at roughly the same time, we might lose a recorded stat, but the end result not much of a harm.

Now if we can register metrics adhoc, if two threads attempt to register the same metric, how does topologyContext handle that?  Under the old metric system trying to register the same metric key twice resulted in an exception.

Crim: Sounds good, looks like you have a good grasp on how this works.  Yeah I'd swap the map implementations out, otherwise looks good.

stanlemon:
After talking with @Crim and looking at how the metrics system feeds into graphite I'm removing the task id prefixing from the `DropwizardRecorder` as it does not seem necessary.

PR #120, opened 2018-07-16
Zookeeper reference implementation cleanup

stanlemon:
Most of this is simple cleanup. There is one notable change though, in the original reference implementation for 0.10 the zookeeper watch supports multiple roots. I don't recall what the reason for this complexity was, probably accounting for the weird way we tried handling global sidelines in the implementation that preceded this. I am confident that there's a better/cleaner way of doing that. I am just not sure that additional complexity is worth it.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/JSON.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/JSON.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2017, 2018, Salesforce.com, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+ * following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this list of conditions and the following
+ *   disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+ *   disclaimer in the documentation and/or other materials provided with the distribution.
+ *
+ * * Neither the name of Salesforce.com nor the names of its contributors may be used to endorse or promote products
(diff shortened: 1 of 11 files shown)


src/main/java/com/salesforce/storm/spout/sideline/recipes/trigger/zookeeper/ZookeeperWatchTrigger.java
```diff
@@ -112,19 +115,11 @@
         .setDateFormat("yyyy-MM-dd HH:mm:ss")
         .create();
 
-    /**
```
stanlemon: I know this seems weird, but I prefer the docs from the interface in situations like this where they're just copy pasted.

src/main/java/com/salesforce/storm/spout/dynamic/JSON.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.gson.GsonBuilder;
+import com.salesforce.storm.spout.sideline.persistence.FilterChainStepTypeAdapterFactory;
+
+/**
+ * Thin wrapper around JSON parsing.
+ *
+ * Intended to hide the particular JSON implementation of the day.
+ */
+@SuppressWarnings("checkstyle:AbbreviationAsWordInName")
+public class JSON {
```
Crim: Json instead of all caps?  Does this pass checkstyle?

stanlemon: See the annotation?

Crim: Seems like you violated checkstyle

stanlemon: Yes. But with malice and forethought.

PR #122, opened 2019-01-03
Fix sideline zk configuration

stanlemon:
The 'root' configuration only supports one root, this typo is a holdover from the old implementation where we supported a list of roots to watch.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/sideline/recipes/trigger/zookeeper/ZookeeperWatchTrigger.java
+++ b/src/main/java/com/salesforce/storm/spout/sideline/recipes/trigger/zookeeper/ZookeeperWatchTrigger.java
@@ -47,10 +47,10 @@ import org.apache.curator.framework.recipes.cache.PathChildrenCacheListener;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.xml.bind.DatatypeConverter;
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
+import java.math.BigInteger;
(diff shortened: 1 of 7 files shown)


Crim:
Should we be compiling the list of changes in the CHANGELOG as part of these PRs?

src/test/java/com/salesforce/storm/spout/dynamic/VirtualSpoutTest.java
```diff
@@ -79,6 +80,14 @@
  */
 public class VirtualSpoutTest {
 
+    /**
+     * Before each test reset the static on the MockConsumer.
+     */
+    @BeforeEach
+    void resetMockConsumerPartitions() {
+        MockConsumer.partitions = Arrays.asList(1);
```
stanlemon: @Crim this ended up being my flapper problem. Some day I'll revisit the `MockConsumer` entirely and fix this in a better way. You called it though. 😉

src/main/java/com/salesforce/storm/spout/sideline/recipes/trigger/zookeeper/ZookeeperWatchTrigger.java
```diff
@@ -327,6 +325,14 @@ private SidelineRequestIdentifier generateSidelineRequestIdentifier(
         return new SidelineRequestIdentifier(identifier.toString());
     }
 
+    private String generateIdFromJson(final String dataJson) throws NoSuchAlgorithmException {
```
stanlemon: Functionally the same as what was there before, just no longer dependent upon javax.

PR #123, opened 2019-01-04
Implement helper for handling trigger events

stanlemon:
The goal here is to provide an implementation for generating the trigger events leveraged by the recipe. This can be used in a variety of places, such as command-line tooling or a thin api.

### DIFF
--- a/src/main/java/com/salesforce/storm/spout/dynamic/Tools.java
+++ b/src/main/java/com/salesforce/storm/spout/dynamic/Tools.java
@@ -28,6 +28,10 @@ package com.salesforce.storm.spout.dynamic;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
 
+import java.math.BigInteger;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.HashMap;
@@ -125,5 +129,22 @@ public class Tools {
             .map(String::trim)
             .filter((inputString) -> !inputString.isEmpty())
             .toArray(String[]::new);
+    }
+
+    /**
+     * Create an MD5 hash from a string.
+     * @param data string to make an md5 hash for.
+     * @return md5 hash.
+     */
+    public static String makeMd5Hash(final String data) {
+        try {
+            // Use the data map, which should be things unique to define this criteria to generate our id
+            final MessageDigest md5 = MessageDigest.getInstance("MD5");
+            md5.update(StandardCharsets.UTF_8.encode(data));
+            final String id = String.format("%032x", new BigInteger(1, md5.digest()));
+            return id;
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


src/test/java/com/salesforce/storm/spout/sideline/recipes/trigger/TriggerEventHelperTest.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package com.salesforce.storm.spout.sideline.recipes.trigger;
+
+/**
+ * Test the {@link TriggerEventHelper} to make sure it starts, resumes and resolves sidelines for the
+ * recipe implementation.
+ */
+class TriggerEventHelperTest {
+    // TODO: Implement tests
```
Crim: i . think something is missing here

src/main/java/com/salesforce/storm/spout/dynamic/Tools.java
```diff
@@ -126,4 +130,21 @@ public static String prettyDuration(Duration duration) {
             .filter((inputString) -> !inputString.isEmpty())
             .toArray(String[]::new);
     }
+
+    /**
+     * Create an MD5 hash from a string.
+     * @param data string to make an md5 hash for.
+     * @return md5 hash.
+     */
+    public static String makeMd5Hash(final String data) {
```
Crim: does this return as a hex string?

src/test/java/com/salesforce/storm/spout/sideline/recipes/trigger/TriggerEventHelperTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Create shared zookeeper test server.
+     */
+    @RegisterExtension
+    static final SharedZookeeperTestResource sharedZookeeperTestResource = new SharedZookeeperTestResource();
+
+    private static Map<String, Object> config = new HashMap<>();
+    private static TriggerEventHelper triggerEventHelper;
+    private static CuratorFramework curator;
```
Crim: Does it make sense to add a close method to CuratorHelper?  If so we can no hold a reference to CuratorFramework here and delegate everything to CuratorHelper....

I haven't investigated closely, so I'll defer to your judgement

Crim: err this comment is for TriggerEventHelper.java :Facepalm:

stanlemon: Thus far `CuratorHelper` has been just a helper, so it's never been used without direct access to the underlying `CuratorFramework` instance. I think my preference is people just close the `CuratorFramework` instance and we keep `CuratorHelper` as lean and precise as possible. Thoughts?

src/main/java/com/salesforce/storm/spout/dynamic/Tools.java
stanlemon: Yes.

src/test/java/com/salesforce/storm/spout/sideline/recipes/trigger/TriggerEventHelperTest.java
stanlemon: Thus far `CuratorHelper` has been just a helper, so it's never been used without direct access to the underlying `CuratorFramework` instance. I think my preference is people just close the `CuratorFramework` instance and we keep `CuratorHelper` as lean and precise as possible. Thoughts?

PR #130, opened 2019-10-31
Remove FilterChainBuilder and serialize the steps directly

stanlemon:


src/main/java/com/salesforce/storm/spout/sideline/recipes/trigger/zookeeper/ZookeeperWatchTrigger.java
```diff
@@ -293,18 +285,17 @@ private SidelineRequest buildSidelineRequest(final TriggerEvent triggerEvent) {
...
      * @return A sideline request identifier for the filter chain step
      * @throws NoSuchAlgorithmException Your java install is whack yo, it's missing MD5, for realz???
      */
     private SidelineRequestIdentifier generateSidelineRequestIdentifier(
-        final TriggerEvent triggerEvent,
-        final FilterChainStep step
+        final TriggerEvent triggerEvent
     ) {
-        final String json = gson.toJson(step);
+        final String data = json.to(triggerEvent.getFilterChainStep());
```
kevpeek: it seems like the `TriggerEvent` should perhaps know how to make its identifier.

stanlemon: Should be resolved now, let me know what you think @kevpeek

